PR #5, opened 2026-05-06
feat(skills): skill management ecosystem + self-improving Jarvis

omnigentx:
## Summary

Branch ships two intertwined features plus the voice-pipeline polish that landed in earlier commits:

1. **Skill management** — full CRUD via dashboard (`/skills` library page + Agent Detail integration), per-agent attach/detach, all hot-reloadable via fast-agent's `rebuild_agent_instruction` (no restart needed for content edits or attach/detach).
2. **Self-improving Jarvis (experimental, opt-in)** — Jarvis gets MCP tools to author/edit/attach/delete skills at runtime. Tools live in a subprocess but mutations route back to the live backend via a generic Unix-domain-socket RPC framework (`services/runtime_rpc.py`) so `rebuild_agent_instruction` runs in the right process. Every Jarvis-driven mutation writes a notification so the user can review.
3. **Voice + UI polish** (older commits on the branch): hands-free conversation loop, plug-and-play STT/TTS, markdown+mermaid renderer, `${VAR}` proxy fix, voice settings panels.

## Notable design choices

- **MCP subprocess + UDS RPC** over in-process tools. Picks up future use cases (sandboxed tool execution, multi-language tools, Jarvis dynamically registering new MCP servers) without rework. Generic framework so future tools register `(method, handler)` pairs alongside the skill ones.
- **Hot-toggle, not hot-attach** for the experimental flag. `skill_server` is always registered; each tool reads `experimental/SELF_IMPROVING_ENABLED` from the DB on every call. Toggle flips behavior the next tool call — no restart.
- **Slim `attach`/`detach` payload** — `{agent, skill, persisted, skill_count}`. The full manifest array (with bodies) used to be returned; cut ~600x token cost per call. Frontend re-fetches separately.
- **Built-in protection** via `_builtin.yaml` manifest. Edits allowed; deletes blocked. Manifest absence fails open (dev-mode safe).
- **Atomic rollback** for card-based attach/detach: YAML write reverted if `rebuild_agent_instruction` fails. Disk and runtime stay consistent.
- **`mtime_ns` as string** in JSON — 19-digit nanoseconds exceed `Number.MAX_SAFE_INTEGER`, would silently lose precision and trigger 409 on every save.

## Tests

- **Backend**: 902 passing. Highlights:
  - `test_skills` (90+) — name validation, CRUD, frontmatter rules, optimistic locking, built-in protection, agent-card cleanup, attach/detach + rollback
[...]

PR #46, opened 2026-05-22
fix(agents): surface real MCP/skill load state in agent details

omnigentx:
## Summary

- Agent details UI was reading from **static config** and always rendering "Connected" — masking the MCP attach failures and missing-skill resolution errors that fast-agent silently swallows. Users were reassured about tools their agents could not actually call.
- This PR makes the panel **faithfully reflect what the live agent runtime sees**: failed MCP servers show a red badge with the attach error in a tooltip; requested-but-unresolved skills show as failed; spawned agents show a pending skeleton instead of fabricating template data while the live introspection is in flight.
- Push-based refresh: `SpawnProgressBridge` broadcasts a new `runtime_config_ready` event over the existing activity stream — **no polling** added.

## Why

The dashboard claimed agents had tools they couldn't use. Across 8 agents in my local backend that's:
- 4–5 "skills" listed but unresolved on Finance/Research/CrawlStories cards (50% silent miss)
- MCP servers shown green when their stdio subprocess never started

The user's review intuition matched: ["nó chỉ show ra cho user yên tâm thôi, chứ thực tế chưa chắc nó đã thấy"](../jarvis-2-workspace).

## What changed

**Backend (`backend/routes/agents.py`):**
- `_get_agent_skills` diffs raw `config.skills` (requested) vs the live agent's `skill_manifests` (loaded). Path-style refs (`".fast-agent/skills/X"`) compared by basename so card-based agents don't get false-positive failed rows.
- `_get_runtime_tools` returns `{server: {tools, status, error}}` including `_configured - _attached` as failed entries.
- `_enrich_mcp_errors_async` (new) populates `error` via `agent.get_server_status()` on the detail endpoint only — keeps the list endpoint cheap.
- `_build_spawn_agent_detail`: drops the team-template fallback for `skills`/`tools`; returns `runtime_pending=true` until the child process emits its `runtime_config` event.
- Helper pure functions: `_parse_skill_names`, `_merge_skill_status`, `_merge_spawn_tool_status`.
- Removed orphans: `_load_role_config_from_template`, `_get_spawn_server_tools` (no callers after fallback path was deleted).

**Backend (`backend/services/spawn_progress_bridge.py`):**
- `_broadcast_runtime_ready` helper. Called from `_handle_runtime_config` and `_handle_mcp_status` after the DB upsert, so connected dashboards refetch push-based.

**Frontend (`dashboard/src/views/AgentDetail.vue`):**
- `serverTools` / `allTools` adapted to the new shape (with legacy `Array` fallback during the deploy window).
- Red badges on failed skills/servers, count chip in panel headers, error banner in the server accordion body.
- Pulse-shimmer skeleton when `runtime_pending`.
- Watches `store.recentEvents` for `runtime_config_ready` matching the current agent → refetch. No polling.
- Instruction tab switched to `content-type="text"` — markdown sanitiser was stripping fast-agent's literal `<available_skills>`, `<skill>`, `<scripts>` template tags, lying about what the LLM actually sees.
- Banner "Defined in agent.py" now keys on `type==='builtin'` (was `!card`); spawned `team`/`dynamic` agents get an accurate "ephemeral, edit team template to persist" message.

**Dev convenience (`dashboard/vite.config.js`):**
- `VITE_PORT` / `VITE_PROXY_TARGET` env overrides (defaults unchanged). Lets a second workspace run side-by-side without editing the file.

## Tests

- `tests/test_routes/test_agents.py`: 16 new tests, 38 total pass. Covers loaded-vs-failed surface, path-style skill normalization, async error enrichment, `runtime_pending=true` when runtime_config absent, merge helpers, and `runtime_config_ready` broadcast.
- All 308 route tests pass.

## Test plan

- [ ] Open Agent Details for a built-in agent → MCP Servers tab shows real attach state; expand a failed server → see attach error.
- [ ] Open Agent Details for a card-based agent (FinanceAgent / ResearchAgent) → no false-positive failed skills despite path-style refs in YAML.
- [ ] Spawn a team agent → before runtime_config event arrives, panels show pulse skeleton; once child emits, panels populate without manual refresh.
- [ ] Open Instruction tab for any agent that uses skills → fast-agent's `<available_skills>` block renders literally instead of being stripped.
- [ ] Open spawned agent's Skills tab → banner says "spawned agent, ephemeral" (not "defined in code (agent.py)").

### DIFF
--- a/backend/routes/agents.py
+++ b/backend/routes/agents.py
@@ -1126,6 +1126,11 @@ def _merge_spawn_tool_status(
     mcp_status.servers (per-server is_connected + error_message) into the
     unified ``{server: {tools, status, error}}`` shape that the built-in agent
     path also returns.
+
+    Note: the ``mcp_servers`` payload emitted by isolated_runner._emit_mcp_status
+    carries BOTH ``status`` ("connected"/"failed") and ``is_connected`` (bool).
+    We read ``is_connected``; the spawn_progress_bridge log path reads ``status``.
+    Both fields are load-bearing — dropping either silently regresses one reader.
     """
     result: dict[str, dict] = {}
     for server_name, tools_list in (rt_tools or {}).items():
--- a/backend/routes/agents.py
+++ b/backend/routes/agents.py
@@ -9,6 +9,7 @@ import re
 import logging
 import sqlite3
 from pathlib import Path
+from typing import Any
 
 import json as _json
 import yaml
@@ -334,18 +335,70 @@ RELOAD_SIGNAL = Path(__file__).parent.parent / ".fast-agent" / ".reload_needed"
 
 
 def _get_agent_skills(agent_name: str) -> list[dict]:
-    """Read resolved skills from FastAgent runtime AgentConfig."""
+    """Return per-skill load status for the agent details UI.
+
+    Each entry carries ``status``:
+      * ``"loaded"`` — manifest was resolved + injected into the live agent.
+      * ``"failed"`` — requested by name in config but missing from the live
+        agent's resolved manifests (file missing, parse error, etc).
+
+    The live agent's ``skill_manifests`` is the authoritative source of what
+    the agent actually sees; the raw ``config.skills`` field (preserved by
+    ``_apply_skills_to_agent_configs``) is the request log used to compute
+    the failed delta.
+    """
     try:
         agent_data = fast.agents.get(agent_name)
         if not agent_data:
             return []
         config = agent_data.get("config")
-        if not config or not hasattr(config, "skill_manifests"):
+        if not config:
             return []
-        return [
-            {"name": m.name, "description": m.description, "content": m.body or ""}
-            for m in config.skill_manifests
+
+        # Loaded manifests — prefer the live agent instance (post-init), fall
+        # back to the resolved config list if the runtime isn't ready yet.
+        loaded: list = []
+        if state.agent_app is not None:
+            instance = state.agent_app.get_agent(agent_name)
+            if instance is not None:
+                loaded = list(getattr(instance, "skill_manifests", None) or [])
+        if not loaded:
+            loaded = list(getattr(config, "skill_manifests", None) or [])
+
+        result = [
+            {
+                "name": m.name,
+                "description": m.description or "",
+                "content": m.body or "",
+                "status": "loaded",
+            }
+            for m in loaded
         ]
+        loaded_names = {m.name for m in loaded}
+
+        # Requested-but-not-loaded — string entries in raw config.skills
+        # may be either bare skill names ("user-context") or directory paths
+        # (".fast-agent/skills/user-context"). Fast-agent resolves both via
+        # SkillRegistry to a SkillManifest whose ``name`` is the directory
+        # basename, so we compare by basename to avoid false-positive
+        # "failed" entries for path-style refs. Non-string entries
+        # (Path/SkillRegistry/SkillManifest) reference directories rather
+        # than a specific skill, so we can't diff them.
+        requested = getattr(config, "skills", None)
+        if isinstance(requested, (list, tuple)):
+            for item in requested:
+                if not isinstance(item, str) or not item:
+                    continue
+                basename = item.rstrip("/").split("/")[-1]
+                if basename in loaded_names:
+                    continue
+                result.append({
+                    "name": basename,
+                    "description": "",
+                    "content": "",
+                    "status": "failed",
+                })
+        return result
     except Exception as e:
         logger.debug("[AGENTS API] Failed to get skills for %s: %s", agent_name, e)
         return []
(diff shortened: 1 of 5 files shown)


omnigentx (commented):
## PR Review (GitNexus-assisted) — Risk: **LOW**

### Scope
5 files, +878 / −206. Reshapes `/api/agents/*` `tools` from `dict[str, list[dict]]` → `dict[str, {tools, status, error}]`, adds per-skill `status: loaded|failed`, adds `runtime_pending` to spawn detail, adds `runtime_config_ready` SSE event. GitNexus impact analysis confirms every changed/removed symbol is private to `backend/routes/agents.py` — zero external callers, list/get endpoints are the only public consumers and both are updated here.

### Findings

**1. ⚠️ LOW — `_build_agent_dict` drops the config-fallback for tools.**
`_get_runtime_tools(name) or dict(getattr(config, "tools", {}) or {})` → just `_get_runtime_tools(name)`. When `state.agent_app` is None (boot window, or the existing test at `test_agents.py:69-73`), the list endpoint now returns `tools={}` instead of the configured server set. Honest behavior, but worth confirming the brief startup empty-state is acceptable — the dashboard relies on the push refresh to repopulate.

**2. ⚠️ LOW — `runtime_config_ready` watcher inspects only `events[0]`.**
`AgentDetail.vue:60-68` reads `store.recentEvents[0]` on each watch fire. Vue 3's `flush: 'post'` coalesces same-tick mutations; if a second event for any agent lands in the same microtask after `runtime_config_ready`, the watch fires once with the newer event at [0] and the refetch is skipped. EventSource normally delivers events as independent microtasks, so this is rare in production — but a head-watermark or `.find(e => e.event_type === 'runtime_config_ready' && e.agent_name === agentName.value)` scan over newly-prepended events would be safer. The pattern at `AgentDetail.vue:91` already does the filter-style read.

**3. ⚠️ LOW — `_handle_mcp_status` broadcast has no unit test.**
`TestRuntimeReadyBroadcast` covers `_handle_runtime_config` (`test_agents.py:902-913`) but the equivalent broadcast added inside `_handle_mcp_status` (`spawn_progress_bridge.py:838-840`) isn't exercised. One mirror test would close the gap.

**4. ⚠️ LOW (soft SSoT) — two readers, two keys.**
`_emit_mcp_status` in `isolated_runner.py:929-937` emits both `status` and `is_connected` on each server. `_merge_spawn_tool_status` reads `is_connected`; the `_handle_mcp_status` log code reads `status`. Co-emitted today, so this works — but if either field is ever dropped, the other reader silently regresses. Worth a one-line comment in `_emit_mcp_status` that both keys are load-bearing.

**5. ℹ️ INFO — local test env couldn't import `fast_agent`.**
`uv run pytest backend/tests/test_routes/test_agents.py` failed on collection (`ModuleNotFoundError: fast_agent`) on `main` as well — not introduced by this PR. The new tests read soundly (clean stubs, clear setup), but I couldn't independently rerun the 38/308 pass claim. Please confirm green in CI.

### Missing Coverage
- Frontend: no unit tests for the new `runtime_pending` skeleton, the legacy-`Array` fallback in `serverTools`, or the SSE watch — PR test plan is manual-only.
- `_handle_mcp_status` broadcast (see #3).

### What's Good
- All d=1 callers of changed/removed symbols live in the same file and are updated in the same diff. `_load_role_config_from_template` and `_get_spawn_server_tools` are cleanly removed (impact analysis confirms no external refs).
- Aggregator / agent fields used by new code (`_attached_server_names`, `_configured_server_names`, `_server_to_tool_map`, `McpAgent.get_server_status`, `ServerStatus.error_message`) all exist where assumed.
- Honest defaults: `runtime_pending: true` + empty arrays instead of synthesizing template data the live agent might not see. Push-based refresh keeps it from blocking the user (no polling added).
- Path-style skill names normalized by basename — directly addresses the false-positive case for card-based agents called out in the description.
- `Instruction` tab switched to `content-type="text"` so fast-agent's literal `<available_skills>` / `<skill>` tags survive markdown sanitization — closes a real UX-lying bug.
- Backwards-compat: `serverTools` computed tolerates both legacy `Array` and new `dict` shape during the deploy window.

### Recommendation
**APPROVE** with optional follow-ups #2 and #3. Nothing here blocks merge.

omnigentx (commented):
## Follow-up review — `eae43c5` addresses prior findings

All three actionable items from the previous review are resolved:

- **#2 Watcher race — FIXED.** `AgentDetail.vue:62-79` now walks newly-prepended events from head until `_lastSeenEvent`. Confirmed safe: `pushEvent` in `stores/agents.js:117` spreads (`[event, ...recentEvents.value].slice(0,50)`) so existing event references are preserved → the `events[i] === _lastSeenEvent` identity check terminates correctly. Same-tick coalescing under `flush: 'post'` no longer drops events.
- **#3 `_handle_mcp_status` broadcast test — FIXED.** `test_handle_mcp_status_broadcasts` mirrors the runtime_config test with realistic per-server payload (one connected, one failed with error). Asserts the broadcast lands with the right `agent_name` / `run_id`.
- **#4 SSoT comment — FIXED.** `_merge_spawn_tool_status` now documents that both `status` and `is_connected` in the `mcp_status` payload are load-bearing for different readers.

Findings #1 (list-endpoint returns `tools={}` when `state.agent_app` is None during boot) and #5 (local pytest env couldn't import `fast_agent`) were informational and don't block merge.

No new issues introduced. **APPROVE.** 👍

PR #48, opened 2026-05-22
feat(pause): full PauseController refactor — instant + correct + team-wide

omnigentx:
## Summary

End-to-end refactor of the agent pause/resume flow. Realizes all five user requirements from the design discussion:

1. **Instant pause** for LLM calls — cancel-and-retry contract via new `on_pause_cancel` hook (no waiting through a 30s LLM stream). Tool calls intentionally cooperative ("strategy B" — side-effects can't be unrolled, let them complete).
2. **Realtime correct UI state** — four SSE events (`agent_pausing` → `agent_paused` → `agent_resuming` → `agent_resumed`) emitted at the right moments. `agent_paused` now comes from *inside* the checkpoint hook the instant the agent is about to `await`, not from the pause endpoint synchronously.
3. **Resume from exact pre-pause state** — the retry path reissues the LLM call with unchanged `_delta_messages`; conversation history is preserved.
4. **Team-wide pause** — pausing any team member expands to the full team via `_resolve_scope`. Orchestrator included (it's just another team member in the registry). Late-joiner hook in `spawn_progress_bridge` auto-pauses agents spawned into a paused team.
5. **Auto-attached hook + centralized API** — single `pause_controller.attach(agent)` replaces the previously scattered "create + double merge_hooks + branch on existing" pattern in `routes/chat.py`, `routes/inject.py`. Idempotent via `_pause_attached` sentinel.

Plus restart recovery: new `agent_pause_state` table persists who's paused; `restore_on_startup()` re-applies on server boot.

## Phases (8 commits)

| # | Phase | What changed |
|---|---|---|
| 1 | Skeleton | `services/pause_controller.py` introduced as drop-in for old `pause_manager.py`. Pure rename; old module becomes a shim re-export. |
| 2 | State machine | Three-state lifecycle + 4 SSE events. Subprocess (submodule) mirrors the same machine via signal handler + checkpoint hook. |
| 3 | LLM cancel | `on_pause_cancel` hook + retry loop in `ToolRunner.__anext__` (submodule). PauseController tracks `_current_tasks` per agent via the `before_llm_call` hook; pause cancels iff in LLM phase. |
| 4 | Team scope | `pause(scope)` accepts agent OR team name; `_resolve_scope` expands. Late-joiner hook in `SpawnProgressBridge._upsert_spawn_record`. `approval_service` refactored to call team-scope pause. |
[...]

PR #10, opened 2026-05-08
feat(auth): cookie+CSRF session — kill the 401-loop, surface AuthGate

omnigentx:
## Summary

Replaces the long-lived API key in `localStorage` + `?api_key=` query auth with a proper httpOnly session cookie + CSRF double-submit, and adds an `AuthGate` modal that blocks the dashboard when the session is invalid. Fixes the original incident from 2026-05-08 where rotating the master key sent the dashboard into a forever-401 loop with nowhere for the user to type the new key.

Four logically-distinct commits — please review them in order, each is self-contained.

## Why this fixes the bug

Before:
- `EventSource` has no status code on `onerror`; every SSE composable retried 401s with exponential backoff *forever* while the user stared at the "Disconnected" badge.
- The API key lived in `localStorage` cleartext (XSS = permanent compromise) and leaked into the URL of every SSE / audio / WS connection (nginx logs, Sentry, browser history).
- Rotating `JARVIS_API_KEY` in Settings did not refresh the in-page session — every subsequent request 401'd until a manual reload.

After:
- Session is a 1h HMAC-signed cookie embedding a fingerprint of the current `JARVIS_API_KEY`. Rotation invalidates every existing session immediately, no server-side state.
- A single `auth` Pinia store centralizes the state machine (`unknown → authenticated → challenged → unauthenticated`). 401s soft-fail through `challenged` and re-probe once before opening the modal — backend restart blips no longer eject the user.
- A pub/sub bus lets the 9 SSE composables and the voice WebSocket tear down on `EXPIRED` and reconnect on `RESTORED` without each one rediscovering auth state through 401s.
- `AuthGate.vue` mounts at root, blocks the entire app while unauthenticated, traps focus, and surfaces the backend's typed `reason` (`key_rotated`, `expired`, `max_lifetime_exceeded`, …) verbatim.
- Settings → rotate key auto-logs-in with the new key after the PUT succeeds, so the user stays on the same page.

## Threat model deltas

| Threat | Before | After |
|---|---|---|
| XSS lifts auth | API key, valid forever | Session cookie httpOnly; not readable by JS. CSRF cookie is readable but useless without a session. |
| URL leakage of auth | `?api_key=` in every SSE / WS URL | Cookie auto-attached; URLs no longer carry secrets |
| Rotation propagation | Manual reload required | Sessions across all tabs invalidate within one HTTP round-trip |
| Cross-tab logout | Per-tab manual | BroadcastChannel multicasts transitions; sibling tabs auto-react |
[...]

PR #49, opened 2026-05-26
feat: passkey login + cookie-only SPA auth (+ PR #48 follow-up)

omnigentx:
## Summary

Three threads that landed together in one PR (12 commits, easy to review individually):

1. **Passkey login** — Touch ID / Face ID / Windows Hello / security keys as the SPA's primary auth path. Per-deployment, scoped to each user's domain; API key in `.env` remains as scripts credential + recovery path.
2. **Cookie-only SPA** — drops the long-standing `localStorage.jarvis_api_key` mirror + Bearer header. SPA authenticates solely via the httpOnly `jarvis_session` cookie + CSRF double-submit. XSS that exfiltrates localStorage no longer lifts the credential. Non-browser callers (Xiaozhi voice device, scripts) keep using Bearer.
3. **PR #48 follow-up** — fixes uncovered while manually testing the merged pause/resume work: UV-PID gotcha (subprocess agents died on pause), pause-cycle UI flicker, terminal "paused" event missing for subprocess path, and tooling polish.

## Why this matters

For the upcoming OSS release: each user self-hosts at their own domain. Passkey makes "open browser → Touch ID → in" the default UX, with `.env` recovery if every passkey is lost. Cookie-only removes the XSS exfiltration footgun the auth-cookie PR (#10) couldn't fully close until callers migrated off Bearer.

## Linked PR

Blocks on **omnigentx/fast-agent#4** (submodule pause fixes). The pointer bump in this PR currently targets that branch's SHA — after #4 merges, a follow-up commit will re-pin to fast-agent's main HEAD.

## Commits (review order)

| # | Commit | Why bundled |
|---|---|---|
| 1 | `fix(pause): PR #48 follow-up` | Leftover work from prior session — UV-PID, paused-cycle UX, monitor utils extraction |
| 2 | `feat(auth): WebAuthn lib + models + wrapper` | Foundation |
| 3 | `feat(auth): passkey HTTP routes` | 7 endpoints + integration tests |
| 4 | `fix(auth/ws): voice WS 3-way auth` | Required so the cookie-only migration doesn't break voice |
| 5 | `feat(setup): wizard step 1 mints session cookie` | Required so the wizard works without localStorage Bearer |
| 6 | `feat(dashboard): passkey UI` | service + SettingsAuth panel |
| 7 | `feat(dashboard/auth): AuthGate passkey button + drop localStorage write` | Both touch AuthGate.vue |
| 8 | `refactor(dashboard): drop localStorage jarvis_api_key` | The migration itself, across 10 files |
| 9 | `fix(vite): forward X-Forwarded-Host` | Without this, dev RP ID = "127.0.0.1" → passkey ceremony fails |
| 10 | `test(e2e): passkey + cookie-only regression specs` | Virtual authenticator harness + 3 specs |
| 11 | `docs: SELF_HOSTING passkey section` | + CLAUDE.md §5 §6 (debug discipline lessons) |
| 12 | `chore(submodule): bump fast-agent` | Pin to fast-agent#4 branch SHA |

## Test plan

Backend (`cd backend && uv run pytest tests/`):

- [x] `test_core/test_webauthn.py` — 27 unit tests
- [x] `test_routes/test_passkey_routes.py` — 28 integration tests
- [x] `test_routes/test_ws_voice_auth.py` — 9 unit tests for 3-way precedence
- [x] `test_routes/test_setup.py` — 25 tests including new cookie-mint assertion
- [x] `test_routes/test_auth_session.py` — existing 19 tests still pass
- [x] **Total auth-related backend: 126/126 pass**

Frontend (`cd dashboard`):

- [x] `npm run test:unit` — 140/140 pass (18 new for `services/passkey.js`)
- [x] `npx playwright test` — 61/65 pass; 4 failures are pre-existing on main (agent-monitor.spec, team-monitor-v2.spec — unrelated to this PR, missing `/api/approvals` in fixtures)

Manual:

- [x] Register passkey from Settings → Authentication → Touch ID → row appears with correct label/transports
- [x] Logout (clear cookies in DevTools) → AuthGate shows "Sign in with passkey" → Touch ID → straight into dashboard, no key typed
- [x] "Use API key instead" fallback still works
- [x] DevTools → Application → Local Storage: **no `jarvis_api_key` key present** after login
- [ ] Setup wizard from scratch (user's open task — instructions in PR conversation: backup `.env` + `data/jarvis.db`, restart backend, walk through wizard, verify cookie minted at step 1)

## Gotchas worth calling out in review

- **AuthGate.vue uses `auth._setAuthenticated` directly** instead of `auth.login`. Justified inline — avoids a redundant `whoami` round-trip after passkey verify, since the finish response already carries the csrf_token + expires_in we'd otherwise re-fetch.
- **Mock-backend `fulfilled` array race**: `waitForRequest` resolves BEFORE the route handler pushes to `fulfilled`; the register E2E uses `waitForResponse` to dodge this. Documented in the spec comment.
- **Vite proxy `changeOrigin: true` is preserved**; we forward `X-Forwarded-Host` instead. Matches what nginx/Caddy do — backend behaves the same in dev and prod.
- **Submodule pointer pins to a branch SHA**, not main. Reachable via `origin/fix/pause-uv-pid-and-path-b`. Will re-pin to main after fast-agent#4 merges.
- **`seedApiKey` / `clearApiKey` E2E helpers are no-ops now** — kept to avoid churning every existing spec; no-op is documented inline.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -461,7 +461,7 @@ trial and error.
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **jarvis-2-workspace** (36513 symbols, 102696 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
 > If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
@@ -477,7 +477,7 @@ This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relatio
 
 1. `gitnexus_query({query: "<error or symptom>"})` — find execution flows related to the issue
 2. `gitnexus_context({name: "<suspect function>"})` — see all callers, callees, and process participation
-3. `READ gitnexus://repo/jarvis/process/{processName}` — trace the full execution flow step by step
+3. `READ gitnexus://repo/jarvis-2-workspace/process/{processName}` — trace the full execution flow step by step
 4. For regressions: `gitnexus_detect_changes({scope: "compare", base_ref: "main"})` — see what your branch changed
 
 ## When Refactoring
@@ -516,10 +516,10 @@ This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relatio
 
 | Resource | Use for |
 |----------|---------|
-| `gitnexus://repo/jarvis/context` | Codebase overview, check index freshness |
-| `gitnexus://repo/jarvis/clusters` | All functional areas |
-| `gitnexus://repo/jarvis/processes` | All execution flows |
-| `gitnexus://repo/jarvis/process/{name}` | Step-by-step execution trace |
+| `gitnexus://repo/jarvis-2-workspace/context` | Codebase overview, check index freshness |
+| `gitnexus://repo/jarvis-2-workspace/clusters` | All functional areas |
+| `gitnexus://repo/jarvis-2-workspace/processes` | All execution flows |
+| `gitnexus://repo/jarvis-2-workspace/process/{name}` | Step-by-step execution trace |
 
 ## Self-Check Before Finishing
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,16 +60,50 @@ For multi-step tasks, state a brief plan:
 
 Strong success criteria let you loop independently. Weak criteria ("make it work") require constant clarification.
 
+## 5. Don't Code from Imagination
+
+Read the actual code + logs before proposing a fix. No guessing what
+a provider/hook/handler "probably does" — open the file, grep the
+source, tail the log. Every fix should cite the file:line that
+proves the bug.
+
+## 6. Happy Path FIRST
+
+**Rule:** every user-facing change ships with ≥1 e2e test that
+exercises the real click-flow through production code. No mocking the
+subsystem being changed. If a user could hit a basic failure on their
+first manual try, the happy-path test was missing.
+
+Trap to refuse: `@pytest.fixture(autouse=True)` that stubs the very
+subsystem under test. That's a regression test dressed up as e2e —
+opt-out (or convert the fixture to opt-in).
+
+### Anti-patterns (auto-reject in review)
+
+- **Silent fallback** when LLM/UI input disagrees with DB ground truth → `raise`, don't guess.
+- **Frontend store with no seed on mount** → cross-store gates render stale on fresh tab.
+- **AF_UNIX path > 104 bytes on macOS** → use `server._short_unix_socket()`.
+- **UI status whitelist not following the backend state machine** → when backend adds states (e.g. pausing/resuming), grep `statusLabel`/`statusColor`/`statusAccent`/`PAUSE_CYCLE` and update them all; "Unknown" / wrong-colored dot is a silent UX bug.
+
+### Pre-PR checklist (pause/approval/spawn changes)
+
+- [ ] Real e2e test (no mock of the changed subsystem) added to the inventory above.
+- [ ] LLM-supplied input: authoritative DB lookup + `raise` on mismatch.
+- [ ] New SSE event: frontend store consumer test asserts reactive UI update.
+- [ ] New persisted state: restart-recovery covered.
+- [ ] New AgentCard gate: parent view's `onMounted` seeds the dependency store.
+- [ ] Click-flow spelled out in PR body.
+
 ---
 
-**These guidelines are working if:** fewer unnecessary changes in diffs, fewer rewrites due to overcomplication, and clarifying questions come before implementation rather than after mistakes.
+**These guidelines are working if:** fewer unnecessary changes in diffs, fewer rewrites due to overcomplication, clarifying questions come before implementation rather than after mistakes, and manual testing turns up no "basic case" failures.
 
 ---
 
 <!-- gitnexus:start -->
 # GitNexus — Code Intelligence
 
-This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
+This project is indexed by GitNexus as **jarvis-2-workspace** (36513 symbols, 102696 relationships, 300 execution flows). Use the GitNexus MCP tools to understand code, assess impact, and navigate safely.
 
 > If any GitNexus tool warns the index is stale, run `npx gitnexus analyze` in terminal first.
 
@@ -85,7 +119,7 @@ This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relatio
 
 1. `gitnexus_query({query: "<error or symptom>"})` — find execution flows related to the issue
 2. `gitnexus_context({name: "<suspect function>"})` — see all callers, callees, and process participation
-3. `READ gitnexus://repo/jarvis/process/{processName}` — trace the full execution flow step by step
+3. `READ gitnexus://repo/jarvis-2-workspace/process/{processName}` — trace the full execution flow step by step
 4. For regressions: `gitnexus_detect_changes({scope: "compare", base_ref: "main"})` — see what your branch changed
 
 ## When Refactoring
@@ -124,10 +158,10 @@ This project is indexed by GitNexus as **jarvis** (36383 symbols, 102213 relatio
 
 | Resource | Use for |
 |----------|---------|
-| `gitnexus://repo/jarvis/context` | Codebase overview, check index freshness |
-| `gitnexus://repo/jarvis/clusters` | All functional areas |
-| `gitnexus://repo/jarvis/processes` | All execution flows |
-| `gitnexus://repo/jarvis/process/{name}` | Step-by-step execution trace |
+| `gitnexus://repo/jarvis-2-workspace/context` | Codebase overview, check index freshness |
+| `gitnexus://repo/jarvis-2-workspace/clusters` | All functional areas |
+| `gitnexus://repo/jarvis-2-workspace/processes` | All execution flows |
+| `gitnexus://repo/jarvis-2-workspace/process/{name}` | Step-by-step execution trace |
 
 ## Self-Check Before Finishing
 
--- a/backend/core/auth_cookies.py
+++ b/backend/core/auth_cookies.py
@@ -1,0 +1,72 @@
+"""Shared cookie-setting helpers for the auth routes.
+
+Lives outside ``routes/auth.py`` so other route modules that need to
+mint or clear the session cookies (the setup wizard, the future passkey
+re-auth path) can import a stable, public symbol instead of reaching
+into ``routes.auth`` for an underscore-prefixed function.
+
+The previous ``routes.auth._set_auth_cookies`` lived in the file that
+*also* owns the login route — fine within ``routes/auth.py``, but a
+cross-module private-symbol import (``from routes.auth import
+_set_auth_cookies``) breaks silently when the source module changes
+its signature. Moving the helper here makes the cookie contract
+intentionally part of the public API surface.
+"""
+from __future__ import annotations
+
+from fastapi import Request, Response
+
+from core.session import CSRF_COOKIE_NAME, SESSION_COOKIE_NAME, SESSION_TTL_SECONDS
+
+
+def is_secure_request(request: Request) -> bool:
+    """Decide whether to mark cookies ``Secure``.
+
+    True when the originating scheme is HTTPS or behind a proxy
+    forwarding ``X-Forwarded-Proto: https``. In dev (``http://localhost``)
+    we leave Secure off so the cookies are actually accepted by the
+    browser.
+    """
+    if request.url.scheme == "https":
+        return True
+    fwd = request.headers.get("x-forwarded-proto", "")
+    return "https" in fwd.lower()
+
+
+def set_auth_cookies(
+    response: Response,
+    request: Request,
+    session_token: str,
+    csrf_token: str,
+) -> None:
+    """Set both auth cookies with consistent attributes.
+
+    Used by ``POST /api/auth/login``, ``POST /api/auth/refresh``,
+    ``POST /api/auth/passkey/authenticate/finish``, and
+    ``POST /api/setup/auth`` — anywhere the backend hands a fresh
+    session to the SPA.
+    """
+    secure = is_secure_request(request)
+    common = {
+        "max_age": SESSION_TTL_SECONDS,
+        "httponly": True,
+        "secure": secure,
+        "samesite": "lax",
+        "path": "/",
+    }
+    response.set_cookie(SESSION_COOKIE_NAME, session_token, **common)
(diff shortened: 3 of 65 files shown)


omnigentx (commented):
## Review summary

Reviewed via local git diff + GitNexus. **Verdict: shippable after addressing 3 HIGH-severity items below.** The cookie-only migration is well-executed (XSS attack surface is meaningfully reduced), the WebAuthn integration follows the spec on every meaningful point I checked, and the PR #48 follow-up correctly applies the SSoT pattern. The HIGH items are not show-stoppers but they undercut the security posture the PR is selling.

---

### 🔴 HIGH

**H1. `require_user_verification=False` weakens primary-credential sign-in** (`backend/core/webauthn.py:259`)

`verify_authentication(...)` passes `require_user_verification=False`. Combined with `user_verification=PREFERRED` in the registration options, a passkey can be presented WITHOUT user verification (no Touch ID / Face ID / FIDO2 PIN — just presence/possession). Modern platform authenticators DO perform UV, but the server never checks the assertion's UV flag. A stolen YubiKey without PIN can authenticate alone.

For a *primary credential* sign-in path, this is too permissive. Recommend either:
- `require_user_verification=True` in both `build_authentication_options` (`UserVerificationRequirement.REQUIRED`) AND `verify_authentication`, or
- explicitly inspect `verified.user_verified` post-verify and reject if False, or
- document the threat model that justifies UV=False (single-user self-hosted + physical access already assumed?).

Same applies to register at `webauthn.py:212`. For register the session is already authenticated so the impact is smaller, but the cleanest stance is REQUIRED everywhere.

**H2. WebSocket cookie auth has no Origin check** (`backend/routes/ws_voice.py:60-95`)

When `_ws_authenticated` accepts the session cookie, no Origin verification happens. Modern browsers DO honor SameSite=Lax for WebSocket upgrades, mitigating CSWSH — but this is recent behavior, browser-implementation-dependent, and silently bypassed by some proxies that strip SameSite attributes. Defense-in-depth: when the cookie path wins, also check `ws.headers.get(\"origin\")` against an allow-list (e.g., derived from `X-Forwarded-Host` or a configured `TRUSTED_ORIGINS` env). Cookies-only auth without Origin verification is the standard CSWSH vector and worth closing.

**H3. `passkey_authenticate_begin` leaks all credential IDs to unauthenticated callers** (`backend/routes/auth.py` passkey_authenticate_begin)

The route's docstring claims:
> We pass an empty `allow_credentials` so the browser uses discoverable-credential UX

But the code passes the real list:
```python
allow_ids = [c.id for c in creds]
ceremony_id, options = wa.build_authentication_options(
    request=request, allow_credential_ids=allow_ids,
)
```

Either the comment is right and the code is wrong, or vice versa — they disagree. Since the credentials were registered with `resident_key=REQUIRED`, the empty-list path works (browser shows discoverable creds). Real harm: `/authenticate/begin` is public; an unauthenticated probe leaks the credential ID list. Recommend passing `[]` and updating tests to assert the response shape carries no allowCredentials.

---

### 🟡 MEDIUM

**M1. Wizard silently breaks on missing JWT_SECRET** (`backend/routes/setup.py:_mint_wizard_session`)

```python
except RuntimeError as exc:
    # JWT_SECRET missing — degrade gracefully. The wizard can still
    # finish via the legacy Bearer path if the FE has the key
```

The Bearer-fallback in the SPA was removed in commit 8 of this PR. So if `JWT_SECRET` is unset:
1. Step 1 returns 200 (looks fine to user)
2. No cookie minted
3. No Bearer sent on step 2 (SPA has no key cached)
4. Step 2 → 401 → user sees \"Authentication required\" with no recovery path

This is the silent-fallback anti-pattern that CLAUDE.md §5 now explicitly bans (\"Silent fallback when LLM/UI input disagrees with DB ground truth → raise\"). Recommend hard-fail in `setup_auth` with 503 + actionable detail (\"set JWT_SECRET in .env and restart\").

**M2. No rate-limit on `/passkey/authenticate/begin` and `/passkey/has-any`**

Both are public, DB-touching, allocating ceremony state. `/begin` adds to the in-process `_ceremonies` dict (capped only by 5-min TTL) — a malicious client can spam it to balloon memory before TTL evicts. Wire the shared `_check_rate_limit(client_ip)` bucket on both.

**M3. Idempotent-insert race on `passkey_register_finish`**

Two simultaneous finish calls for the same physical authenticator race past the \"existing is None\" check and both attempt `db.add(...)` — the second hits UniqueConstraintViolation as a 500. Wrap with `try/except IntegrityError → refresh row` to match the existing \"replaced\" branch's intent.

**M4. `_set_auth_cookies` cross-module from private symbol** (`setup.py` imports from `routes.auth`)

`_set_auth_cookies` (underscore-prefixed) is now called from both `routes/auth.py` and `routes/setup.py`. Cross-module private-symbol imports break silently when the source module changes its signature. Lift the helper to a shared module (`core/auth_cookies.py` or expose it as `set_auth_cookies` public from `routes.auth`).

**M5. UV-PID fix: \"pause-immediately-after-spawn\" race** (`pause_controller.py:_find_python_child`)

`_find_python_child` returns `uv_pid` as fallback when `pgrep -P` finds zero children — but that's exactly the bug this PR is fixing (SIGUSR1 → uv → TERMINATE → agent dies). The fallback re-introduces the bug for the race window where uv hasn't yet forked python. Either:
- retry the pgrep walk with a short backoff (50ms × 5) before falling back, or
- refuse the pause with a clear error: \"agent still spawning, retry in a moment\".

The current behavior is \"fail safer than before, but still potentially fatal in a small window\".

**M6. `_ws_authenticated` is wedged between imports** (`ws_voice.py:56-95`)

The function definition sits between `from services import shared_state` and `from services.sse_progress import ...`, and references `logger` which is defined at line 115. Functionally works (Python lazy attr resolution), but breaks PEP 8 (all imports → all module globals → all functions). Move below `logger = logging.getLogger(__name__)`.

---

### 🟢 LOW / nits

- **L1.** `auth._setAuthenticated` is private (underscore) but now legitimately called from `AuthGate.handlePasskey`. Promote to public (`setAuthenticated`).
- **L2.** `_find_python_child` falls through to `children[0]` silently when no python is found. Log a warning so the operator can see \"signaled non-python child, may not handle SIGUSR1\".
- **L3.** `logger.warning(\"[AUTH] passkey register verify failed: %s\", exc)` — `exc` may carry attestation bytes. Use `exc.__class__.__name__` or `type(exc).__name__` to avoid log leakage.
- **L4.** Two consecutive `_db = SessionLocal()` blocks in `approval_service.py` — merge into one session.
- **L5.** CLAUDE.md GitNexus repo name renamed to `jarvis-2-workspace`. If this is regenerated by `npx gitnexus analyze`, the diff is just churn — but fresh clones with a different repo dir name will see a different value. Worth a one-liner on whether the rename should be checked in.

---

### What's good (worth calling out)

- **Cross-flow defence in `pop_ceremony`** (register kind vs authenticate kind) — exactly the right shape, prevents ceremony hijacking across flows.
- **`expected_user_id` binding in `verify_registration`** even in single-user mode — future-proofs multi-user migration.
- **Sign_count bumped AFTER verify** (`auth.py:passkey_authenticate_finish`) — correct ordering, comment explains why.
- **Cookie attrs**: `httponly=True`, `samesite=lax`, conditional `secure` based on `_is_secure_request` (which honors `X-Forwarded-Proto`). Clean.
- **CSRF middleware design**: exempt list is minimal and well-justified; pass-through when no CSRF cookie matches the docstring's stated scope.
- **Approval `team_name` authoritative DB lookup** is exactly the SSoT pattern CLAUDE.md §9 asks for. Replaces the silent fallback that hid the original incident.
- **`upsertAgentPreservingPauseCycle`** centralizes the \"who owns the status field\" decision — one helper, every handler consults it. SSoT pattern on FE.
- **E2E tripwire spec** (`auth-cookie-only.spec.ts`) covers all three leak vectors: localStorage, Bearer header, URL `?api_key=`. The kind of test that pays for itself the first time someone naively re-adds `localStorage.setItem(...)`.
- **UV-PID diagnosis** in `_find_python_child` is correct + the WARNING comment in fast-agent's `isolated_spawner.py` (PR #4) is the matching durable signpost.

---

### Suggested action

Merge after H1-H3 are addressed (either fix or explicitly accepted with a comment). M-items can land as follow-ups but are worth filing as issues. The PR description's gotchas list is accurate; the test plan is thorough; the commit-by-commit review order works.

Linked PR review: omnigentx/fast-agent#4 has separate findings (Path A/B asymmetry, missing hook-ordering test).

omnigentx:
Addressed in [c8c305d](https://github.com/omnigentx/jarvis/pull/49/commits/c8c305d). All HIGH + MEDIUM + LOW items applied. 179/179 backend tests + 140/140 dashboard unit tests pass locally; new tests added for every behavioral change.

### HIGH (merge blockers)

**H1 — UV REQUIRED on register + authenticate.** ``require_user_verification=True`` in the lib call, ``UserVerificationRequirement.REQUIRED`` in both options. Documented threat model inline (stolen YubiKey-without-PIN is no longer sufficient — primary credential needs Touch ID / PIN / Face ID, not just possession).

**H2 — WebSocket Origin check on cookie path.** New ``_ws_origin_allowed(ws)`` derives the deployment origin from request headers (X-Forwarded-Host honored to match production reverse-proxy setups) and rejects mismatches. Bearer + query paths bypass the Origin check because they're credential-of-possession that browsers cannot present cross-origin. ``TRUSTED_WS_ORIGINS`` env extends the allow-list for multi-front-end setups. 4 new tests cover matching/missing/wrong Origin and the Bearer-bypass interaction.

**H3 — empty allow_credentials in /authenticate/begin.** Route always passes ``[]`` now. Docstring + test name spell out the rationale: discoverable credentials work via the resident-key registration; echoing IDs to unauthenticated callers pre-staged a device fingerprint. ``allow_credential_ids=[]`` is now a deliberate constant, not a doc-vs-code split.

### MEDIUM

**M1 — wizard 503 on JWT_SECRET missing.** ``_mint_wizard_session`` raises 503 with actionable detail ("set JWT_SECRET in .env and restart"). The Bearer fallback that previously hid this was removed by [16f81a4](https://github.com/omnigentx/jarvis/pull/49/commits/16f81a4) — without M1 the operator would see step 2 401 with no recovery path. New ``test_missing_jwt_secret_raises_503`` covers it. Also fixes the CI ``test-backend`` job failure (CI had no JWT_SECRET, hit the silent-degrade path).

**M2 — rate-limit on /passkey/has-any + /passkey/authenticate/begin.** Both share the per-IP /login bucket (5/60s). Without this a malicious client could spam either to balloon DB load or in-process ceremony state. 2 new burst-trip tests.

**M3 — IntegrityError race fix on register/finish.** Wrapped insert in ``try/except IntegrityError → re-query + update``. Same shape as the "existing is not None" fast path, just resolves the commit-time race. New ``test_race_winner_loses_to_existing_row_returns_replaced_true`` pins the user-visible contract.

**M4 — ``core/auth_cookies.py``.** ``_set_auth_cookies`` + ``_is_secure_request`` + ``_clear_auth_cookies`` lifted to public symbols in a new module. ``routes/setup.py`` imports from there; ``routes/auth.py`` keeps thin shim aliases for internal callers. No more cross-module private-symbol import.

**M5 — UV-PID race window.** ``_find_python_child`` retries 5x with 50ms backoff (covers the spawn race), then returns ``None`` on exhaustion instead of falling back to uv_pid. The fallback re-introduced the bug — signaling uv → SIGUSR1 default action TERMINATE → uv dies → python orphaned. Test renamed + reasserts the new "fail safe" behavior. Also added L2: warning log when pgrep finds non-python child fallback ("signaled non-python child, may not handle SIGUSR1").

**M6 — ``_ws_authenticated`` placement.** Moved below ``logger = ...`` definition. Imports → module globals → functions per PEP 8.

### LOW

* **L1**: ``_setAuthenticated`` → ``setAuthenticated`` (auth store + 3 call sites).
* **L2**: covered in M5 above.
* **L3**: passkey verify exception logs ``type(exc).__name__`` instead of ``str(exc)`` — InvalidRegistrationResponse may carry raw attestation bytes.
* **L4**: ``approval_service.create_request`` merged the two ``SessionLocal()`` blocks into one shared session.

### Test budget

| Layer | Count | Result |
|---|---|---|
| Backend (auth/setup/passkey/ws-auth/pause/approval) | 179 | ✅ all pass |
| Dashboard unit | 140 | ✅ all pass |
| Dashboard E2E | 61/65 | 4 failures are pre-existing on main (agent-monitor / team-monitor v2 — missing /api/approvals in fixture; unrelated to this PR) |

### CI fix

The previous ``test-backend`` CI failure was ``test_setup_auth_mints_session_cookie_so_step2_works_without_bearer``. Root cause: CI runner had no JWT_SECRET, so the wizard's session mint silently no-op'd, leaving the cookie unset. The M1 change makes this state fail loud everywhere; the test's autouse fixture now sets JWT_SECRET so the happy path stays exercised.

### Items intentionally left in this PR

* **L5 — CLAUDE.md GitNexus rename** stays. ``npx gitnexus analyze`` regenerates the block on every run with the current repo dir name; the rename was the user's actual local clone path. Worth a separate convention doc if you want repo-dir-name independence.

omnigentx (commented):
Verified all items addressed in [c8c305d](https://github.com/omnigentx/jarvis/pull/49/commits/c8c305d) — diffs match the response comment. Spot-checks below.

**HIGH**

- **H1** REQUIRED at all four callsites (``build_registration_options``, ``verify_registration``, ``build_authentication_options``, ``verify_authentication``). Each has an inline why-comment naming the threat model (stolen YubiKey-without-PIN). Library will raise on UV-unset assertion → 401 fail-closed.
- **H2** ``_ws_origin_allowed`` mirrors ``origin_from_request`` in webauthn (honors X-Forwarded-Host / X-Forwarded-Proto), and crucially the cookie-rejected-Origin path *falls through* to Bearer/query instead of short-circuiting — confirmed by ``test_session_cookie_wrong_origin_but_valid_bearer_accepted``. ``TRUSTED_WS_ORIGINS`` extension covered too.
- **H3** ``_passkey_authenticate_begin_impl`` no longer queries the DB at all for credentials; ``allow_credential_ids=[]`` is a deliberate constant with the rationale inline. The doc/code split that was the original concern is gone.

**MEDIUM / LOW**

- M1 raises 503 with actionable detail ("set JWT_SECRET in .env and restart") — closes the silent-wedge path that the cookie-only migration created.
- M2 ``/has-any`` + ``/authenticate/begin`` both go through ``_check_rate_limit`` + ``record_login_attempt`` against the shared per-IP bucket.
- M3 ``_update_existing`` helper extracted; ``IntegrityError`` rollback + re-query handles the race winner case without 500s.
- M4 ``core/auth_cookies.py`` is the new public seam; ``routes/auth.py`` keeps thin underscored aliases for internal callers — clean migration that doesn't churn unrelated call sites.
- M5 5×50ms retry then ``return None`` (no fallback to uv_pid) — the fallback was exactly the bug. L2's "non-python child" warning log is here too.
- M6 helpers below ``logger = ...``; PEP 8 ordering restored.
- L1/L3/L4 all picked up.

**CI follow-up** ([fc22d9f](https://github.com/omnigentx/jarvis/pull/49/commits/fc22d9f)): JWT_SECRET defaulted in conftest is the right shape — M1's hard-fail requires every test path to either set the env or assert the 503. ``/api/approvals`` in shared fixtures resolves the pre-existing 4 E2E failures called out in the original test plan.

Nothing further from me. Ready to merge.

PR #51, opened 2026-05-28
feat(voice): Soniox realtime STT + TTS + chat dictation mic

omnigentx:
## Summary

Two related changes, separate commits for review:

1. **Soniox realtime STT + TTS as plug-and-play voice provider** (`aaf89da`)
   - New WebSocket clients for `wss://stt-rt.soniox.com/transcribe-websocket` and `wss://tts-rt.soniox.com/tts-websocket`.
   - Registry-driven — Soniox surfaces automatically in Settings → Voice with 28 voices + 15 ISO languages (vi + en first).
   - Soniox shares one `api_key` slot across TTS and STT (`voice.secrets.soniox.api_key`) — voice_config + routes/voice now look up engines in either registry so a user picking only Soniox STT can still set the key from the STT card.
   - Pulled the WS-streaming scaffolding (thread + queue + hook + reconnect) into a `WSStreamingSTT` base class so the next cloud STT (Deepgram/AssemblyAI) is a sibling-file change.

2. **Bottom-mic dictation in ChatInput** (`d4f52ff`)
   - Replaces MediaRecorder→WebM-attach with press-to-talk transcription.
   - Backend accepts `{type:'start', mode:'dictation'}` on `/ws/voice`; gates `_dispatch_user_turn` so LLM/TTS stays cold while transcripts still propagate.
   - New `useDictationSession` composable disjoint from `useVoiceSession` (the hands-free conversation pipeline).
   - User speaks → text streams into the textarea → user edits → user hits Send. The hands-free VoiceBar's auto-submit behaviour is untouched.

## Test plan

Verified locally on `feat/soniox-realtime-tts-stt`:

- [x] Backend unit tests — 762 service + route tests pass
  - [x] `test_voice_engine_registry.py` (new `TestSonioxBackend` class)
  - [x] `test_soniox_stt.py` — 11 token-protocol assertions (config message, `<end>` endpoint flush, provisional/final accumulation)
  - [x] `test_soniox_tts.py` — 6 provider tests
  - [x] `test_voice_routes.py::test_stt_only_engine_secret_surfaces`
  - [x] `test_ws_voice.py::test_dictation_mode_suppresses_llm_dispatch` + negative-control `test_default_mode_still_dispatches_llm_on_final_transcript`
- [x] Backend e2e against real loopback `websockets.serve()` — `test_soniox_realtime_flow.py` (5 tests: STT roundtrip with `<end>` flush, STT server error, TTS roundtrip, PCM format swap, TTS quota error)
- [x] Frontend Playwright suite — 49/49 pass including
  - [x] `settings-voice-soniox.spec.ts` — pick Soniox in both panels → save API key → pills flip → Save changes posts correct body
  - [x] `chat-input-dictation.spec.ts` — mic click → mocked WS pushes transcripts → textarea fills → user edits → Send fires edited text
- [x] Manual smoke
  - [x] Open `localhost:3001/settings` → Voice → Soniox TTS card lists 28 voices (Adrian default) + 15 language dropdown (en, vi first)
  - [x] Soniox STT card renders secret slot with "set api_key" pill until key is saved
  - [x] Live Soniox API call (requires real API key) — not exercised in CI

## Notes for reviewer

- WS_URL is a `ClassVar` on `WSStreamingSTT` — patching `SONIOX_STT_WS_URL` module constant won't reach it; tests must patch the class attribute (`monkeypatch.setattr(soniox_stt.SonioxSTTService, "WS_URL", ...)`).
- The dictation backend gate is a local var in `voice_ws`, read by `on_stt_event` and assigned by the receive loop in the same function scope — no `nonlocal` needed.
- `dashboard/` is left untouched; `frontend/` is the canonical UI per `fdb6be5`.

### DIFF
--- a/backend/routes/ws_voice.py
+++ b/backend/routes/ws_voice.py
@@ -284,6 +284,14 @@ async def voice_ws(ws: WebSocket) -> None:
     tts_task: Optional[asyncio.Task] = None
     user_query_pending = False  # set when STT finalises text but agent task hasn't started yet
     bot_speaking = False        # True from tts_start to tts_end / interruption
+    # Dictation mode: client picked the bottom mic in ChatInput (press-to-talk
+    # transcription) instead of the top hands-free button. In this mode the
+    # STT pipeline runs as usual but ``final_transcript`` does NOT trigger
+    # the LLM agent turn — the client is using transcripts to populate a
+    # text input the user will edit + submit manually. The flag is flipped
+    # by ``{"type":"start","mode":"dictation"}`` and never resets within a
+    # session (one socket = one role).
+    dictation_mode = False
     # AEC diagnostic — when bot is speaking, we copy incoming mic PCM into
     # this buffer + track RMS so we can prove whether echoCancellation is
     # actually scrubbing the loudspeaker bleed. Dumped to a WAV at tts_end.
@@ -346,7 +354,14 @@ async def voice_ws(ws: WebSocket) -> None:
         # Mark a user turn as ready to dispatch on the asyncio side. The
         # actual scheduling happens in the main receive loop so we stay on
         # the event-loop thread for create_task.
-        if name == "final_transcript":
+        #
+        # Dictation mode short-circuits here: the client wants the raw
+        # transcript (already emitted via the standard ``final_transcript``
+        # event a few lines up) so it can put it in a text box for the user
+        # to edit. Skipping ``_dispatch_user_turn`` keeps the LLM + TTS
+        # pipeline cold, which is the whole point — no wasted tokens, no
+        # bot voice talking back.
+        if name == "final_transcript" and not dictation_mode:
             text = (payload or {}).get("text") or ""
             if text.strip():
                 user_query_pending = True
@@ -755,7 +770,14 @@ async def voice_ws(ws: WebSocket) -> None:
                     logger.info("[ws_voice diag] frontend report: %s", payload)
                 elif kind == "stop":
                     return
-                # "start" is a noop — accepting the socket already starts
+                elif kind == "start":
+                    # Accepting the socket already starts the STT pipeline; the
+                    # client sends ``start`` mostly as a handshake-ack. The
+                    # one piece of payload we honour is ``mode: "dictation"``
+                    # which gates LLM/TTS dispatch on ``final_transcript``.
+                    if payload.get("mode") == "dictation":
+                        dictation_mode = True
+                        logger.info("[ws_voice] dictation mode enabled for this session")
     except WebSocketDisconnect:
         pass
     finally:
--- a/backend/routes/ws_voice.py
+++ b/backend/routes/ws_voice.py
@@ -353,15 +353,11 @@ async def voice_ws(ws: WebSocket) -> None:
 
         # Mark a user turn as ready to dispatch on the asyncio side. The
         # actual scheduling happens in the main receive loop so we stay on
-        # the event-loop thread for create_task.
-        #
-        # Dictation mode short-circuits here: the client wants the raw
-        # transcript (already emitted via the standard ``final_transcript``
-        # event a few lines up) so it can put it in a text box for the user
-        # to edit. Skipping ``_dispatch_user_turn`` keeps the LLM + TTS
-        # pipeline cold, which is the whole point — no wasted tokens, no
-        # bot voice talking back.
-        if name == "final_transcript" and not dictation_mode:
+        # the event-loop thread for create_task. The dictation gate lives
+        # inside ``_dispatch_user_turn`` (single-threaded — asyncio loop
+        # only) so the worker thread never reads the flag concurrently
+        # with the receive loop's writes.
+        if name == "final_transcript":
             text = (payload or {}).get("text") or ""
             if text.strip():
                 user_query_pending = True
@@ -371,6 +367,16 @@ async def voice_ws(ws: WebSocket) -> None:
         """Fire-and-forget: schedule a coroutine that handles one full turn."""
         nonlocal agent_task, user_query_pending
         user_query_pending = False
+        # Dictation gate. ``dictation_mode`` is set by the receive loop
+        # (also asyncio thread) when the client sends
+        # ``{type:'start', mode:'dictation'}``. Checking here keeps both
+        # the read and the write on the event-loop thread, which is the
+        # only way to make the gate race-free without a lock — the client
+        # wants the raw transcript (already emitted upstream) so it can
+        # populate a text box; skipping the LLM/TTS pipeline below is the
+        # whole point of the flag.
+        if dictation_mode:
+            return
         # Cancel any previous in-flight agent/tts so a barge-in mid-turn
         # cleanly resets to the new user input.
         if (agent_task is not None and not agent_task.done()) or (
--- a/backend/services/stt_backends/soniox.py
+++ b/backend/services/stt_backends/soniox.py
@@ -1,0 +1,148 @@
+"""Soniox real-time STT backend.
+
+Streams 16 kHz int16 PCM (fed via ``feed_audio``) to
+``wss://stt-rt.soniox.com/transcribe-websocket`` and forwards transcripts
+through the same hook surface as :class:`services.stt_realtime.RealtimeSTTService`.
+
+The thread/queue/reconnect/hook scaffolding lives in
+:mod:`services.stt_backends._ws_streaming`. This module only encodes the
+Soniox-specific protocol bits (config-message shape, ``<end>`` endpoint
+token, error format) so adding the next cloud STT (Deepgram, AssemblyAI,
+Google) is a sibling file with the same shape.
+
+Endpoint detection: with ``enable_endpoint_detection=true`` Soniox emits a
+special ``{"text": "<end>", "is_final": true}`` token whenever the model
+decides the speaker stopped. We flush the accumulated finalized tokens as
+the next ``final_transcript`` and reset.
+"""
+from __future__ import annotations
+
+import logging
+from typing import Any, ClassVar
+
+from services.stt_backends._ws_streaming import WSStreamingSTT
+
+logger = logging.getLogger(__name__)
+
+SONIOX_STT_WS_URL = "wss://stt-rt.soniox.com/transcribe-websocket"
+ENDPOINT_TOKEN_TEXT = "<end>"
+
+
+class SonioxSTTService(WSStreamingSTT):
+    """Soniox-flavoured cloud STT over the shared WS streaming base."""
+
+    WS_URL: ClassVar[str] = SONIOX_STT_WS_URL
+    LOG_TAG: ClassVar[str] = "Soniox STT"
+
+    def __init__(self, *, api_key: str, params: dict[str, Any]) -> None:
+        if not api_key:
+            raise ValueError("Soniox STT requires an API key")
+        super().__init__(sample_rate=int(params.get("sample_rate") or 16000))
+
+        self._api_key = api_key
+        self._model = params.get("model") or "stt-rt-v4"
+        self._language_hints = _parse_language_hints(params.get("language_hints"))
+        self._enable_language_id = bool(params.get("enable_language_identification"))
+        self._enable_diarization = bool(params.get("enable_speaker_diarization"))
+
+        # Per-utterance buffer of finalized tokens. Reset on every endpoint.
+        self._final_buffer: list[str] = []
+        self._provisional_tail: str = ""
+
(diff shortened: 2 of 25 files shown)


omnigentx (commented):
## Review — Soniox + dictation

**Risk: MEDIUM.** Architecture is clean: `WSStreamingSTT` base class, single-secret-slot shared across TTS+STT, surgical dictation gate. Test coverage is strong — real `websockets.serve()` loopback for the cloud STT/TTS roundtrip (not mock-only) and a `/ws/voice` TestClient test for the dictation gate with a negative-control twin. Approving with nits.

### Actionable before merge (inline comments)
1. **Dictation gate has a worker-thread race.** Cleaner to read the mode on the main thread or pass it at upgrade.
2. **Provisional-token collapse** when one Soniox frame contains multiple non-final tokens.
3. **Two read paths for `voice.secrets.soniox.api_key`** — STT bypasses `get_engine_secrets`.

### Nits / follow-ups
4. `_ws_streaming._sender` can deadlock-leak a thread if the receiver exits first — `asyncio.wait(..., return_when=FIRST_COMPLETED)` instead of `gather`, or sentinel-push from the receiver on exit.
5. `isRecording` excludes the `connecting` state → button visual lags for a few hundred ms between click and `listening`; `toggleRecording` handles double-clicks correctly but the user can't tell.
6. Cross-registry `tts.get(engine) or stt.get(engine)` silently picks TTS on name collision. Contract today is "same name → same secrets list" by convention — a startup-time assertion or registry-shape unit test would harden it.

_Reviewed via /gitnexus-pr-review against `main`._

backend/routes/ws_voice.py
```diff
@@ -346,7 +354,14 @@ def on_stt_event(name: str, payload: dict[str, Any]) -> None:
...
         # the event-loop thread for create_task.
-        if name == "final_transcript":
+        #
+        # Dictation mode short-circuits here: the client wants the raw
+        # transcript (already emitted via the standard ``final_transcript``
+        # event a few lines up) so it can put it in a text box for the user
+        # to edit. Skipping ``_dispatch_user_turn`` keeps the LLM + TTS
+        # pipeline cold, which is the whole point — no wasted tokens, no
+        # bot voice talking back.
+        if name == "final_transcript" and not dictation_mode:
```
omnigentx: **MEDIUM — worker-thread race on `dictation_mode`.**

`dictation_mode` is a `voice_ws` local set on the asyncio loop by the receive loop (line ~777) but read here from the STT worker thread. The unit test (`test_dictation_mode_suppresses_llm_dispatch`) even compensates for this race by draining a `partial_transcript` first — comment in the test says _"without this short event, the dictation flag race could let final_transcript reach the dispatcher first"_. That acknowledgement is a code smell.

Practical exposure is small (final_transcript fires hundreds of ms into a recording, long after the `start` handshake), but a cleaner design exists:
- **(a)** Pass `mode` as a query parameter at WS upgrade so the flag is set _before_ the receive loop runs, **or**
- **(b)** Move the gate to the main-thread side: always `loop.call_soon_threadsafe(_dispatch_user_turn, text)` and check `dictation_mode` inside the dispatcher, where it's single-writer.

Either removes the cross-thread read and lets you delete the partial-transcript "beat" trick from the test.

backend/services/stt_backends/soniox.py
```diff
@@ -0,0 +1,148 @@
...
+            self._emit_endpoint()
+            return
+
+        if is_final:
+            self._final_buffer.append(text)
+            self._provisional_tail = ""
+        else:
+            # Soniox emits provisional tokens that may be revised; track only
+            # the latest non-final text so the running partial doesn't snowball.
+            self._provisional_tail = text
```
omnigentx: **MEDIUM — provisional tokens collapse if Soniox packs multiple per frame.**

Every non-final token assigns `self._provisional_tail = text`, so a frame containing `[{"text": "hello ", "is_final": False}, {"text": "world", "is_final": False}]` yields only `"world"`. The unit tests use one-provisional-per-frame, but Soniox's wire format can pack multiple tokens per message — the same `_handle_event` loop iterates `data.get("tokens") or []` for finals (appending), so the protocol clearly allows multi-token messages.

Either:
- Confirm via the Soniox spec that provisional tokens are always one-per-frame and add a one-line code comment citing it, **or**
- Track provisional accumulation per-frame: reset `_provisional_tail` at the start of `_handle_event` for provisional-only frames and append within the loop.

backend/services/stt_backends/soniox.py
```diff
@@ -0,0 +1,148 @@
...
+    """Factory used by :mod:`services.stt_realtime._BACKEND_FACTORIES`.
+
+    The API key is loaded from the same encrypted ``voice.secrets.soniox.api_key``
+    slot that the TTS engine uses — Soniox issues one key per account that
+    works against both STT and TTS endpoints, so duplicating slots would
+    only invite drift.
+    """
+    from services.config_service import config_service as _cs
+
+    api_key = _cs.get("voice", "secrets.soniox.api_key")
```
omnigentx: **MEDIUM — SSoT: two read paths for the same secret slot.**

This reads `config_service.get("voice", "secrets.soniox.api_key")` directly; the TTS side goes through `services/voice_config.py:get_engine_secrets`. Both converge today, but the STT bypass means future changes to the helper (hot-reload, decryption tweaks, in-process cache) won't apply to STT — a classic SSoT-by-convention rather than SSoT-by-construction.

Suggested change:
```python
from services.voice_config import get_engine_secrets
api_key = get_engine_secrets(_cs, "soniox").get("api_key")
```
Keeps a single read path for the slot, matching the TTS provider.

frontend/src/components/chat/ChatInput.vue
```diff
@@ -21,12 +26,20 @@ const emit = defineEmits(['send', 'stop'])
...
-const mediaRecorder = ref(null)
-const audioChunks = ref([])
 const fileInputRef = ref(null)
 const { isMobile } = useBreakpoint()
 
+// Dictation state. ``isRecording`` mirrors the session's "actively
+// listening" status so the existing pulse-animation styling keeps working
+// without refactoring the button class binding.
+const dictation = useDictationSession()
+const isRecording = computed(() => dictation.status.value === 'listening' || dictation.status.value === 'loading_stt')
```
omnigentx: **LOW (UX nit) — `connecting` is not visually `isRecording`.**

For the few hundred ms between mic click and `listening`, the button shows the idle state. `toggleRecording` handles a double-click during `connecting` correctly as a stop, but the visual is inconsistent — the user can't tell their click registered.

Trivial fix:
```js
const isRecording = computed(() =>
  ['connecting', 'loading_stt', 'listening'].includes(dictation.status.value)
)
```

omnigentx: Fixed in 398890f — `isRecording` is now `['connecting', 'loading_stt', 'listening'].includes(status)` so the pulse fires from the click, not after the WS handshake.

backend/services/stt_backends/soniox.py
omnigentx: Fixed in 398890f — `build()` now reads through `voice_config.get_engine_secrets(_cs, "soniox")` so both STT and TTS share one decryption path.

backend/services/stt_backends/soniox.py
omnigentx: Fixed in 398890f. `_provisional_tail` is now cleared at the top of `_handle_event` and accumulated within the per-token loop, with the partial emit moved out of `_handle_token` to the post-loop site. Added `test_multiple_provisional_tokens_in_one_frame_concatenate` to lock the behavior in.

backend/routes/ws_voice.py
omnigentx: Fixed in 398890f via option (b) — moved the gate into `_dispatch_user_turn` so both the read of `dictation_mode` and the receive-loop's write run on the asyncio loop. The STT worker thread no longer touches the flag, and the test's partial_transcript "beat" workaround is gone.

omnigentx:
## Round-2 review — `398890f` + `edd5a38` + `1c5c37d`

### `398890f` ✓ — all four review findings landed cleanly

| Finding | Status | Verification |
| --- | --- | --- |
| #1 Dictation gate race | **Fully fixed** | Gate moved into `_dispatch_user_turn` (asyncio thread) — both read and write of `dictation_mode` now serialize through the loop. The test's `partial_transcript` "beat" trick correctly removed. |
| #2 Provisional collapse | **Fully fixed + regression test** | `_provisional_tail` resets at `_handle_event` entry and accumulates within the loop; partial emit moved to one post-loop site. Traced through all 5 existing token tests — still pass under new logic. |
| #3 SSoT secret read | **Fully fixed** | `get_engine_secrets(_cs, "soniox").get("api_key")` — both Soniox halves share one read path. |
| #5 UX `connecting` state | **Fully fixed** | `isRecording = computed(() => ['connecting', 'loading_stt', 'listening'].includes(...))`. |

Bonus observation on #2: the new design emits **one** partial per frame instead of one per token — observable behavior change for clients but strictly better (partials are full-replace, fewer events).

### `edd5a38` — YouTube autoplay (LOW concern)

`?autoplay=1` fires on every iframe mount, not just for freshly-emitted songs. If a user scrolls back to a long session with prior `[[[PLAY:...]]]` tags, every embed in the DOM attempts autoplay on render. Browsers will block most without an attached gesture context, but it's noisy iframe/network activity. The documented happy path (agent emits `[[[PLAY]]]` immediately after the user's typed message → gesture present → autoplay works) is genuinely improved. Worth knowing for future history-scroll behavior.

### `1c5c37d` — sticky save footer

Fine. Two soft nits:
- `z-index: 5` — confirm no popover in the panel (voice-name dropdown, secret reveal toggle, etc.) renders below 5; a `<select>` opening underneath the sticky band would look broken.
- iOS Safari + soft keyboard occasionally double-counts `bottom: 0` against the visual viewport. `dvh` units or `env(safe-area-inset-bottom)` are the usual hardening if the panel is ever reachable in mobile portrait.

### Verdict

Fix-up commit is exactly what I asked for, with a regression test and clear commentary. **APPROVE.**

_Re-reviewed via /gitnexus-pr-review._

omnigentx:
Round-2 nits addressed in 06868d7:

**YouTube autoplay scope** — `youtubeEmbedUrl` now tracks rendered videoIds in a module-scoped `Set`. First sighting per session = `?autoplay=1`; subsequent (history scrollback, conversation re-mount) = `?autoplay=0`. New `autoplayIfFresh` option for opt-out contexts (preview cards, tests). 3 new unit tests pin: first-sight=1, second-sight=0, distinct-ids each get =1, opt-out forces =0 without poisoning the cache.

**Sticky footer z-index** — bumped 5 → 30 so native `<select>` popovers and reveal-eye tooltips render above the save band.

**iOS safe-area** — added `padding-bottom: max(4px, env(safe-area-inset-bottom))` so the home-indicator gutter doesn't swallow the button on iOS portrait.

142 unit + 4 affected Playwright specs pass.

omnigentx:
## Round-3 check — `06868d7`

**Sticky footer hardening ✓**
- z-index 5 → 30, comment cites `SettingsGeneral.vue` parity — good consistency reasoning
- `padding-bottom: max(4px, env(safe-area-inset-bottom))` is the right iOS hardening; desktop's `env()` resolves to 0 so no regression

**YouTube autoplay scope ✓ with one residual edge case**

The session-scoped `Set` solves the **scroll-back** case cleanly: rendering the same `[[[PLAY:...]]]` tag twice in one tab lifetime yields autoplay=0 on the second mount. Opt-out is non-poisoning (good — the test pins that explicitly).

Residual issue: **first load of a thread that already contains historical PLAY tags**. Every embed is "first sight in this tab" → every iframe loads with `?autoplay=1`. Browsers will block them all (no gesture context on fresh tab load) so audio stays silent, but you still pay the iframe-loaded-with-autoplay-query cost — exactly what the round-2 fix was trying to avoid for scrollback. The fix only helps within a single tab session after one full thread render has occurred.

Two ways to close the gap if it matters:
- **Caller-side**: `ChatMessages.vue` passes `{ autoplayIfFresh: isLatestMessage }` so only the bottom-of-thread newly-arrived message qualifies. Older messages always get =0.
- **Prime the cache**: bulk-add every PLAY id from existing message history to `_seenVideoIds` on initial thread mount, before any embed renders. Only new agent emits after that point trigger autoplay.

Not blocking — the practical effect is "extra ?autoplay=1 query for embeds that won't play anyway." Filing as a follow-up.

**Test coverage on the new path ✓** — 3 new unit tests pin first-sight/second-sight/distinct-ids/opt-out behavior. Good regression net.

LGTM, ready to merge from my side.

PR #19, opened 2026-05-15
chore: bump mcp-atlassian — close 2026-05-16 retest gaps

omnigentx:
## Summary

Submodule pointer bump for `backend/mcp-atlassian` (`1cc16ef → 25b25c1`) to pick up two production-retest fixes from omnigentx/mcp-atlassian#2.

## Context

Production retest on 2026-05-16 00:32-00:34 ICT after PR #18 merged. Logs (`jarvis.log:3296,3322-3330`) showed:

1. **Jira `jira_create_project`**: 4 attempts on project JLP, all `HTTP 400: Invalid request payload`. Root cause: Cloud requires `projectTemplateKey` but tool described it as optional → agent passed `""` → generic 400 with no actionable detail.
2. **Confluence `confluence_create_page`**: 403 "permission denied" on space JLP that didn't exist. Confluence returns 403 (not 404) for missing spaces → agent stopped retrying instead of creating the space first.

## Upstream changes (mcp-atlassian#2)

- `ProjectsMixin.create_project`: smart default per `project_type_key` for `projectTemplateKey`. 400 errors now log + surface sent payload + raw response.
- `create_page`: permission-error text rewriter that hints at `confluence_create_space(space_key=..., ...)`.
- Tool descriptions updated.

## Test plan

- [x] mcp-atlassian unit tests: 58 pass (6 new — 3 mixin + 3 server). See submodule PR for details.
- [ ] **Manual on-the-wire verification**: re-trigger the JLP create flow after merge — expect (a) Jira project JLP created with default Scrum template, (b) Confluence page in JLP fails with actionable hint if space not pre-created.

## Merge order

Merge omnigentx/mcp-atlassian#2 first → rebase this PR if needed → merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #18, opened 2026-05-15
fix(mcp,spawn): env expansion + auto-resume SSoT + MCP atlassian tools

omnigentx:
## Summary

Bundle of three independently-rooted issues uncovered while reviewing notification #24 (toolset self-audit) and the 2026-05-15 5612e8f3 retro-meeting deadlock. Spans two submodules + jarvis-side regression tests.

### fast-agent submodule (PR: omnigentx/fast-agent#1)

1. **MCP env ``\${VAR}`` expansion** — ``mcp_connection_manager.py`` now expands placeholders in ``MCPServerSettings.env``. Unblocks ``approval-server`` / ``skill_server`` / ``mcp_admin`` (all use ``\$JARVIS_RUNTIME_RPC_SOCKET``). Evidence: ``jarvis.log.3:18481`` \`[Errno 2]\` on literal placeholder path.
2. **Auto-resume SSoT (Layers B+C)** — ``_check_and_resume_on_inbox`` now hydrates ``env_vars`` from ``record.original_config[\"env_vars\"]`` when caller passes None (Layer B), and derives canonical ``messages_dir`` from ``record.session_id`` + ``project_dir`` when env_vars still lacks ``TEAM_MESSAGES_DIR`` (Layer C). Closes the silent-bail mode behind the 2026-05-15 ``5612e8f3`` deadlock.
3. **send_to_pm signature fix** — auto-wake call was passing ``project_dir`` (kwarg not in signature) and omitting required ``agent_name``. TypeError was swallowed by broad except → PM never auto-woke from this path. Now passes correct kwargs.

### mcp-atlassian submodule (PR: omnigentx/mcp-atlassian#1)

4. **``jira_get_my_account_id``** — new read-only MCP tool; returns the calling user's accountId.
5. **``jira_create_project.lead_account_id``** — accepts ``\"me\"`` sentinel or email/display-name (auto-resolved), not just literal accountId.
6. **``confluence_create_space``** — new write MCP tool wrapping ``ConfluenceFetcher.create_space`` (public/private + best-effort description).

### jarvis-side

7. New regression tests in ``tests/test_services/test_auto_wake_session_path.py`` pinning Layer B (env_vars hydration from record) and Layer C (session_id-derived canonical path). Existing tests untouched.

## Production-incident evidence

- ``jarvis.log.3:18481`` (2026-05-10): ``[Errno 2]`` on literal ``\${JARVIS_RUNTIME_RPC_SOCKET}`` — proves env expansion was missing.
[...]

Issue #17, opened 2026-05-15
Path A inject — propagate run_id across MessageBus → InboxWatcherHook boundary


omnigentx:
## Context

Surfaced by reviewer on PR #16. Token persistence hook reads
``current_run_id`` ContextVar at LLM-call time. Inject Path A
(``_inject_via_message_bus``) sets the ContextVar in the HTTP
request task, but the actual LLM call fires in a DIFFERENT
asyncio task — the alive agent's inbox-watcher loop (running
in the spawned subprocess). ContextVar values do NOT propagate
across task boundaries that way.

Effect: every token row for inject-Path-A turns gets
``run_id=\"\"`` (the empty-string fallback baked into the token
hook). Dashboard token totals continue to drift from 9Router for
every Path A inject — exactly the SSoT divergence pattern PR #16
was supposed to close.

## Why not fixed in PR #16

The proper fix needs cross-package coordination:

1. ``MessageBus.send`` already accepts a ``context: dict`` field
   (line 122 of ``message_bus.py``). Path A would pass
   ``context={\"jarvis_run_id\": request_id}``.
2. ``InboxWatcherHook.before_llm_call`` (in fast-agent submodule)
   would need to set the ContextVar to ``message.context.get(
   \"jarvis_run_id\")`` before ``runner.append_messages(...)``.
3. But ``current_run_id`` ContextVar lives in jarvis's
   ``services/sse_progress.py``. fast-agent submodule can't
   import jarvis modules → architectural violation.

Clean fixes require one of:
- Move the ContextVar to fast-agent as a generic ``current_request_id``
  ContextVar; jarvis imports + uses it.
- Add a ``register_inbox_message_hook(callback)`` mechanism in
  fast-agent that jarvis wires at subprocess startup.
- Add a thread-/task-local registry that fast-agent looks up at
  inject time.

Either way: scope creeps into the submodule API.

## Workaround already in place

- Token row IS written (with empty ``run_id``) — not silently lost.
- Comment in ``routes/inject.py::_inject_via_message_bus``
  documents the gap and links this issue.

## Acceptance criteria

- [ ] Add ``context_meta`` propagation through ``MessageBus`` →
      ``InboxWatcherHook`` → token hook.
- [ ] Path A inject token rows carry ``run_id`` matching the
      original inject request id.
- [ ] Integration test: POST inject Path A → query
      ``SELECT run_id FROM token_usage WHERE agent_name=X`` →
      assert correlated rows present.
[...]

PR #16, opened 2026-05-15
feat: orchestrator + team + meeting stabilization (+ token tracking, inject wake, monitor v2)

omnigentx:
## Summary

Single PR consolidating ~5,000 LOC of work across orchestrator,
team/meeting orchestration, token observability, inject UX, and
monitor UI — all fixing real production incidents from the past
~1 week. Includes 1 submodule bump (fast-agent) + 5 logical
commits on the main repo. User chose option (b) "1 big PR" over
the proposed 4-PR split.

### Why 1 PR

The author prefers a single PR over splits to avoid accidental
code mixing across PR boundaries. Reviewer should walk the
commits in order; each commit message has full scope + rationale.

### Base dependency

This branch is cut from `feat/auth-silent-refresh` (PR #15). The
first 2 commits in the GitHub diff are from PR #15 (`feat(auth):
silent session refresh` + `fix(auth): address PR #15 review`) and
should NOT be reviewed here — they're already covered by PR #15.
After PR #15 merges, this PR auto-rebases and shows only the 5
new commits.

## Commits in order (review walkthrough)

| # | Commit | Scope | LOC |
|---|---|---|---|
| 1 | `feat(orchestrator): team + meeting + auto-wake stabilization` | Submodule bump (fast-agent) + backend integration: orchestrator team_completion correctness, context_persistence mirrors last assistant text into spawn_registry.result, delete_team cascades into notifications, agile_team gains self-audit-tools skill for all 7 roles, spawn_event_socket single-instance guard, agent_message_stream service (new) feeding monitor v2. | ~1500 |
| 2 | `feat(observability): centralize token persistence as always-on hook` | Token tracking moves from per-caller hook attach to startup-time default hook. ContextVar propagates run_id from chat/voice/inject/cron callers. Closes the gap where scheduler-triggered LLM calls bypassed `token_usage` SQLite insert (9Router showed 864K input; Jarvis dashboard 78.8K). | ~480 |
| 3 | `fix(inject): wake alive agent + broadcast started in Path A` | Inject Path A (MessageBus, alive agent) now sends a wake signal via AgentChannel socket + broadcasts `started` activity event for instant UI feedback. Mirrors Path B / Path C contract. | ~170 |
[...]

PR #57, opened 2026-05-31
OSS hardening: 13 fixes from pre-release security review

omnigentx:
## Summary

Pre-release security + bug review produced 32 verified findings (out of 61 candidates). After per-finding discussion with the user, 14 ship in this PR; 5 were skipped with documented reasoning. Each fix lives in its own commit so reviewers can land them piecemeal if needed.

## Fixes

| # | Severity | Title | Commit |
|---|---|---|---|
| #04 | high | `hmac.compare_digest` × 5 sites | `ea87b60` |
| #10 | high | RedactSecretsFilter for all log handlers | `01ac9f2` |
| #09 | high | Drop `docker.sock` mount from both compose files | `2d4dbcf` |
| #02 | high | `safe_story_path` helper × 4 path-join sites | `360a361` |
| #06 | high | Stream + 10 MB cap per chapter (warn at 5 MB) | `8a9163b` |
| #08 | high | Close XSS in MarkdownRenderer (code-chrome + mermaid err) | `4cf7b2b` |
| #01 + #05 | crit + high | Approval gate for `mcp_install` + cron first fire | `cd6efc8` |
| #12 | high | Minor-pin frontend Dockerfile base images | `e284640` |
| #13 | high | Root `NOTICE` for Apache-2.0 attributions | `87e6d73` |
| #14 | high | `SKILL_LICENSING.md` policy for skills sans LICENSE | `3c9e385` |
| #15 | med | `safe_500` helper for 5xx routes (10 sites) | `063d71a` |
| #18 | med | Privacy section + Cloud/Local chip in Voice settings | `6249df7` |
| #19 | med | Split Gipformer (MIT) from sherpa-onnx (Apache-2.0) | `7bdf1dd` |
| submodules | — | Bump fast-agent + mcp-atlassian | `01cd401` |

## Skipped (explicit decision)

| # | Reason |
|---|---|
| #03 SSRF | crawl response → file → TTS, no exfil channel |
| #07 CORS `*`+creds | browser spec blocks the combo + CSRF middleware in place |
| #11 curl\|gpg devcontainer | fixed upstream in `omnigentx/mcp-atlassian#4`, submodule pointer bumped |
| #16 gmail draft validation | drafts only — user reviews `To:` before manual send |
| #17 nginx as root | static file server, breaking change for compose port mapping |

## New tests

* `test_helpers/test_logging_filters.py` — 17 cases pinning redaction patterns
* `test_helpers/test_path_safety.py` — 13 cases (separators, traversal, symlink escape, Vietnamese Unicode happy path)
* `test_helpers/test_http_safety.py` — 5 cases (cap / warn / status pass-through)
* `test_services/test_approval_gate.py` — 10 cases (hash memory, pending coalesce, timeout)
* `frontend/tests/e2e/flows/chat-markdown-xss.spec.ts` — Playwright regression for the lang-injection vector

All backend unit + integration tests pass locally (auth × 39, routes × 52, services full × 790).

## Submodule bumps included

* **fast-agent** → `d6a7ef01` — adds `CHANGES.md` for Apache-2.0 §4(b)
  ([omnigentx/fast-agent#7](https://github.com/omnigentx/fast-agent/pull/7))
* **mcp-atlassian** → `a038a36` — drops `.devcontainer/` curl\|gpg and 6 jira/confluence prod fixes
  ([omnigentx/mcp-atlassian#4](https://github.com/omnigentx/mcp-atlassian/pull/4))

## Test plan

- [ ] `cd backend && uv run pytest tests/test_routes/test_auth.py tests/test_routes/test_auth_session.py tests/test_routes/test_ws_voice_auth.py` — auth + WS auth (39 tests)
- [ ] `cd backend && uv run pytest tests/test_helpers/ tests/test_services/test_approval_gate.py` — new helpers + gate (45 tests)
- [ ] `cd backend && uv run pytest tests/test_routes/` — full routes regression
- [ ] `cd frontend && npm run test:unit` — frontend unit tests (142 cases)
- [ ] `cd frontend && npx playwright test chat-markdown-xss` — XSS regression (requires dev server)
- [ ] Manual: open Settings → Voice → confirm each engine card shows a Cloud (amber) or Local (green) chip
- [ ] Manual: trigger `mcp_install_dependencies` for a new MCP server — verify approval prompt appears with `requirements.txt` content + "use at your own risk" warning
- [ ] Manual: create a cron with `agent_turn` exec mode — verify first fire prompts for approval; subsequent fires of same payload skip the prompt

## Follow-ups (not in this PR)

1. `omnigentx/RealtimeSTT` / `RealtimeTTS` / `figma-ui-mcp` — MIT, no §4(b) compliance work needed.
2. Background-task `uv.lock` regeneration left untracked locally — not part of this security scope.

### DIFF
--- a/backend/helpers/logging_filters.py
+++ b/backend/helpers/logging_filters.py
@@ -70,6 +70,13 @@ class RedactSecretsFilter(logging.Filter):
     Apply to handlers (not loggers) so the scrub catches records propagated
     from child loggers too. On any internal error the record passes through
     unchanged — log delivery is more important than perfect redaction.
+
+    NOTE: this filter only touches ``record.msg`` / merged args. It does
+    NOT see the rendered exception traceback emitted by handlers that use
+    ``exc_info=True`` — that text is produced by
+    :py:meth:`logging.Formatter.formatException` and bypasses the filter
+    chain. Pair this filter with :class:`RedactingFormatter` on every
+    handler to scrub the traceback too.
     """
 
     def filter(self, record: logging.LogRecord) -> bool:  # noqa: A003
@@ -85,3 +92,27 @@ class RedactSecretsFilter(logging.Filter):
             # Logging path must never raise. Best-effort only.
             pass
         return True
+
+
+class RedactingFormatter(logging.Formatter):
+    """Formatter that scrubs the *fully rendered* log line, including the
+    exception traceback that :class:`RedactSecretsFilter` cannot reach.
+
+    ``Logger.error("X", exc_info=True)`` and ``logger.exception(...)`` cause
+    the formatter to append a traceback produced by
+    :py:meth:`logging.Formatter.formatException`. If a secret is stuffed
+    in a frame's local repr or in an exception message, the filter chain
+    has already finished and the secret lands in the file untouched.
+
+    Attaching this formatter to every handler closes that gap by running
+    ``redact_secrets`` over the final output. Never raises — on regex
+    failure it falls back to the unredacted superclass output.
+    """
+
+    def format(self, record: logging.LogRecord) -> str:
+        text = super().format(record)
+        try:
+            return redact_secrets(text)
+        except Exception:
+            # Same rule as the filter: logging path must never raise.
+            return text
--- a/backend/helpers/logging_filters.py
+++ b/backend/helpers/logging_filters.py
@@ -1,0 +1,87 @@
+"""Logging filters — defensive scrub for secret-looking values.
+
+Attached to every handler by ``core.logging_config.setup_logging`` so that
+any log line touching the root logger (or any of the named child loggers
+configured there) has secret-like substrings replaced with ``***``.
+
+This is defense-in-depth: the codebase tries not to log secrets in the
+first place, but a future contributor adding ``logger.debug("config=%s",
+config)`` where ``config`` contains an api_key would otherwise leak it
+to ``logs/jarvis.log`` (and any future APM that ships log lines off-box).
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+
+# Keywords whose value should be redacted in log lines.
+_SECRET_KEYS = (
+    "api_key", "api-key", "apikey",
+    "access_token", "access-token",
+    "auth_token", "auth-token",
+    "secret_key", "secret-key",
+    "password", "passwd",
+    "secret",
+    "token",
+)
+
+# "Authorization" gets its own pattern: the value may be multi-word
+# ("Bearer <jwt>", "Basic <b64>") which the generic KV value class would
+# truncate at the first whitespace.
+_AUTHZ_RE = re.compile(
+    r"(?i)(\bauthorization\s*[\'\"]?\s*[:=]\s*[\'\"]?)"
+    r"([^\'\",}\])]+)"
+)
+
+# Generic key=value / "key": "value" for keywords whose value is a single
+# token. Stops at whitespace / common JSON-dict closers / quotes so the
+# surrounding structure stays intact.
+_KV_RE = re.compile(
+    r"(?i)(\b(?:" + "|".join(re.escape(k) for k in _SECRET_KEYS) + r")\b"
+    r"\s*[\'\"]?\s*[:=]\s*[\'\"]?)"
+    r"([^\s\'\",}\])]+)"
+)
+
+# Standalone "Bearer <token>" anywhere in the line (when not preceded by
+# "Authorization:" — that case is handled by _AUTHZ_RE first).
+_BEARER_RE = re.compile(r"(?i)(\bbearer\s+)(\S+)")
+
+
+def redact_secrets(text: str) -> str:
+    """Replace secret-looking values inside ``text`` with ``***``.
+
+    Idempotent and safe for log lines, JSON snippets, env-style assignments,
+    and ``Authorization: Bearer <token>`` headers. Never raises.
+    """
+    if not text:
+        return text
+    # Apply Authorization first so the multi-word value (Bearer + token)
+    # is consumed before the standalone Bearer/KV patterns run.
+    text = _AUTHZ_RE.sub(r"\1***", text)
+    text = _BEARER_RE.sub(r"\1***", text)
+    text = _KV_RE.sub(r"\1***", text)
+    return text
+
+
+class RedactSecretsFilter(logging.Filter):
+    """Logging filter that scrubs secret-looking values from every record.
+
+    Apply to handlers (not loggers) so the scrub catches records propagated
+    from child loggers too. On any internal error the record passes through
+    unchanged — log delivery is more important than perfect redaction.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:  # noqa: A003
+        try:
+            formatted = record.getMessage()
+            redacted = redact_secrets(formatted)
+            if redacted != formatted:
+                # ``record.getMessage()`` already merged %-args; clear args so
+                # downstream formatters don't try to merge them a second time.
+                record.msg = redacted
+                record.args = None
+        except Exception:
+            # Logging path must never raise. Best-effort only.
+            pass
+        return True
--- a/backend/services/approval_gate.py
+++ b/backend/services/approval_gate.py
@@ -119,6 +119,21 @@ async def gate(
     Returns:
         (approved, reason) — ``reason`` is a short string for logging /
         propagating back to the caller's error envelope.
+
+    Trade-off:
+        Timeout persists the row as ``rejected`` (see ``except
+        asyncio.TimeoutError`` below). Combined with the
+        ``_find_prior_decision`` short-circuit at the top of this
+        function, that means a single missed approval window — eg
+        the operator was AFK for the full ``timeout_s`` — permanently
+        blocks this exact ``(approval_type, scope_key, content_hash)``
+        triple. Every later call returns ``(False, "previously
+        rejected ...")``` without re-prompting. This is intentional:
+        a never-answered cron should not re-spawn a fresh approval
+        every 5 minutes forever. If the operator wants to revisit a
+        timed-out decision, they delete the rejected row (or change
+        the payload so the content_hash differs) and the next call
+        will create a fresh prompt.
     """
     # Local import keeps services/__init__.py free of an import cycle —
     # approval_service depends on activity_stream_manager which depends on
--- a/backend/services/approval_gate.py
+++ b/backend/services/approval_gate.py
@@ -1,0 +1,184 @@
+"""Human-approval gate for high-blast-radius server-side actions.
+
+Used by:
+* :func:`services.mcp_admin_service.install_dependencies` — block
+  ``pip install`` against an LLM-editable ``requirements.txt`` until the
+  user OKs the package list.
+* :func:`services.cron_scheduler.CronScheduler._execute_agent_turn` —
+  block the first execution of any new ``exec_payload`` so a poisoned
+  cron entry can't fire unsupervised at 03:00.
+
+The gate piggybacks on the existing approval pipeline
+(:mod:`approval_service`): create an ApprovalRequestModel, wait for the
+user to decide via the dashboard. Decisions are remembered by
+``(approval_type, scope_key, content_hash)`` so a repeat run with the
+same content auto-proceeds without prompting again. A rejected hash
+stays rejected — re-asking on every cron fire would be noise.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import json
+import logging
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+# Default: wait up to 1 hour for user. After that, treat as rejected so the
+# caller (cron tick, tool call) returns control rather than hanging forever.
+DEFAULT_GATE_TIMEOUT_S = 3600.0
+
+
+def _content_hash(content: str) -> str:
+    return hashlib.sha256(content.encode("utf-8")).hexdigest()[:16]
+
+
+def _matches_scope(row, scope_key: str, content_hash: str) -> bool:
+    try:
+        meta = json.loads(row.metadata_json or "{}")
+    except (json.JSONDecodeError, TypeError):
+        return False
+    return meta.get("scope_key") == scope_key and meta.get("content_hash") == content_hash
+
+
+def _find_prior_decision(
+    approval_type: str, scope_key: str, content_hash: str
+) -> Optional[str]:
+    """Return ``'approved'`` / ``'rejected'`` if a *resolved* approval for
+    this exact (type, scope_key, content_hash) exists. ``None`` otherwise."""
+    from core.database import ApprovalRequestModel, SessionLocal
+
+    db = SessionLocal()
+    try:
+        rows = (
+            db.query(ApprovalRequestModel)
+            .filter(
+                ApprovalRequestModel.approval_type == approval_type,
+                ApprovalRequestModel.status.in_(("approved", "rejected")),
+            )
+            .order_by(ApprovalRequestModel.resolved_at.desc())
+            .all()
+        )
+        for row in rows:
+            if _matches_scope(row, scope_key, content_hash):
+                return row.status
+    finally:
+        db.close()
+    return None
+
+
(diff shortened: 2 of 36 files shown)


backend/services/cron_scheduler.py
```diff
@@ -364,6 +372,38 @@ async def _execute_agent_turn(self, job: CronJobModel) -> str:
...
+            f"**Payload to run:**\n\n```text\n{(payload or '').rstrip()}\n```\n\n"
+            f"---\n\n"
+            f"⚠️ **Use at your own risk.** Approving lets this exact payload "
+            f"run on the schedule above, unsupervised. The target agent has "
+            f"its full tool set available — including any that touch the "
+            f"filesystem, shell, network, or external accounts.\n\n"
+            f"Approve only if you recognise the payload and trust the source "
+            f"that created this job."
+        )
+        approved, reason = await _gate(
```
omnigentx, line 545: **[HIGH] This blocks the entire cron scheduler for up to 1 hour.**

`start()` is a single serial loop that `await`s `_execute_job` → `_execute_agent_turn` → this `await _gate(...)`, and `DEFAULT_GATE_TIMEOUT_S = 3600` (`approval_gate.py:31`). While one `agent_turn` job waits for human approval:

- No other cron job is evaluated or fires — **all schedules freeze** behind one pending approval.
- The DB session opened at `start()` (line ~93) stays open for the whole wait; the run row (line ~150) and `job.status="running"` (line ~159) sit `running` for up to an hour, so the dashboard shows a stuck job.
- `_catch_up_missed` (line ~544) is also serial on one shared session, so several missed `agent_turn` jobs needing approval serialize, each up to an hour.

This is an availability regression introduced by the security fix. Suggested fix: don't block the loop — create the approval and return immediately, then resume/re-fire on resolution via the existing `_signal_resolution` event; or dispatch `_execute_job` as its own task so the loop keeps advancing. Also drop the gate timeout to minutes for this path.

backend/services/cron_scheduler.py
```diff
@@ -364,6 +372,38 @@ async def _execute_agent_turn(self, job: CronJobModel) -> str:
...
+            agent_name=agent_name,
+        )
+        if not approved:
+            logger.warning("[CRON] Job %s blocked by approval gate: %s", job.id, reason)
+            self._broadcast_event("agent_turn_blocked", {
+                "job_id": job.id,
+                "job_name": job.name,
+                "reason": reason,
+            })
+            return f"Approval gate blocked execution: {reason}"
```
omnigentx: **[LOW] A gate-blocked run is recorded as `success`.** This string is returned normally, so `_execute_job` marks the run `success` and emits an `agent_result` notification containing the block text. A job that never actually executed shows up green. Consider a distinct blocked/skipped status (and broadcast `agent_turn_blocked`, which you already emit, as the terminal state).

backend/services/approval_gate.py
```diff
@@ -0,0 +1,184 @@
...
+        resolved = await asyncio.wait_for(
+            approval_service.wait_for_resolution(approval_id),
+            timeout=timeout_s,
+        )
+    except asyncio.TimeoutError:
+        logger.warning(
+            "[GATE] %s/%s — timeout after %.0fs",
+            approval_type, scope_key, timeout_s,
+        )
+        return False, f"approval timeout ({timeout_s:.0f}s)"
```
omnigentx, line 214: **[MEDIUM] Timeout never persists a decision → repeated re-blocking.** On timeout the function returns `(False, …)` but the DB row stays `pending`. On the next fire, `_find_prior_decision` finds nothing resolved and `_find_pending_match` (line ~144) re-attaches to the same stale `pending` row and waits another full `timeout_s`. So a never-answered cron re-freezes the scheduler on every fire (compounds the cron-scheduler finding) and pending rows accumulate. Consider auto-rejecting (or expiring) the row on timeout, or fast-skipping when a sufficiently old pending row already exists.

backend/helpers/logging_filters.py
```diff
@@ -0,0 +1,87 @@
...
+    """Logging filter that scrubs secret-looking values from every record.
+
+    Apply to handlers (not loggers) so the scrub catches records propagated
+    from child loggers too. On any internal error the record passes through
+    unchanged — log delivery is more important than perfect redaction.
+    """
+
+    def filter(self, record: logging.LogRecord) -> bool:  # noqa: A003
+        try:
+            formatted = record.getMessage()
```
omnigentx, line 84: **[LOW] Redaction misses exception tracebacks.** `filter()` only rewrites `record.getMessage()`. Secrets inside `exc_info` tracebacks (local-var reprs, exception messages) are rendered separately by the formatter's `formatException` and bypass this filter — and `safe_500` logs with `exc_info=True`. Also, handlers attached after `setup_logging()` runs (e.g. uvicorn) aren't covered. Reasonable to ship as defense-in-depth, but it doesn't fully match the "scrub every log line" claim — worth a note in the docstring or a follow-up.

omnigentx (commented):
## Review summary — overall solid, one blocker

Most fixes are genuine, correctly wired into the real code paths, and well tested. I verified the wiring directly in the working tree:

- **#04** `hmac.compare_digest` — live at `core/auth.py:107-135`, `routes/auth.py:151`, `routes/ws_voice.py:177-183`; no dead imports.
- **#02** `safe_story_path` — sound logic (per-part reject **and** realpath-under-base check catches symlink escape); wired at all 4 sites; tests cover traversal + symlink + Vietnamese Unicode.
- **#08** XSS — both `innerHTML` sinks replaced with DOM-API/`textContent`; main render path is `DOMPurify.sanitize`'d; Playwright regression present. Good fix.
- **#10** redaction — confirmed live (`setup_logging()` is called at `server.py:20`).
- **#01/#05** approval gate — wired into the real MCP-install (`install_dependencies` ← RPC handler ← `mcp_admin_server`) and cron (`_execute_agent_turn`) paths. Gate internals are correct (event-driven `wait_for_resolution`, race-safe, hash-keyed memory).

The blocker is **not** the gate logic itself but how the cron path consumes it — see the inline comment on `cron_scheduler.py`.

### Additional findings (no inline anchor — unchanged/spread lines)

**[LOW] `get_capped_text` is not applied to every fetch in `crawl_poller`.** The 2000-chapter loop is capped, but `services/crawl_poller.py:201` (provider verification) and `:253` (story-name fetch) still use raw `requests.get(...).text` with no cap, so a single hostile start page can stream unbounded into RAM. Minor completeness gap vs the stated intent of #06.

**[NIT] `_find_prior_decision` / `_find_pending_match` scan all approvals of a type in Python** over `metadata_json` with no index. Fine at current scale; flag for later.

### Test coverage gap
New unit tests are good, but the gate is only exercised in isolation (`test_approval_gate.py`). There is no integration test for gate-inside-scheduler, which is exactly where the blocker below lives.

> Note: posting as a `COMMENT` review (GitHub doesn't allow `REQUEST_CHANGES` on your own PR), but my recommendation is **request-changes** for the cron-scheduler blocker (and ideally the timeout-persistence issue) before merge. Everything else is fine to land.

omnigentx:
Addressed the review with 3 follow-up commits:

- **ab91755** `fix(cron+gate)` — non-blocking dispatch (`asyncio.create_task` + `_inflight_tasks` per job id), new `ApprovalBlocked` exception so `_execute_job` records `run.status="blocked"` + dedicated `job_blocked` event + `blocked` notification type (no fail_count bump), and gate timeout now calls `resolve_approval(..., "reject")` so subsequent fires fast-skip via `_find_prior_decision`.
- **992b18b** `security(logging)` — new `RedactingFormatter` overrides `format()` to scrub the fully rendered line, including the traceback emitted by `formatException`. Wired in `setup_logging` alongside the existing `RedactSecretsFilter` (defense in depth).
- **7676f06** `security(crawl)` — `crawl_poller.py:201` (provider verification) and `:253` (story-title fetch) now go through `get_capped_text` too, closing the completeness gap in #06.

Tests:
- `test_approval_gate.py` — 2 new: timeout persists rejection, timeout reason survives a resolve-failure race.
- `test_cron_scheduler_agent_turn.py` — autouse fixture patches the gate (so the existing `agent_name` regression test still exercises post-gate behaviour); new test asserting `ApprovalBlocked` is raised and `resume_and_send` is never called when the gate refuses.
- `test_logging_filters.py` — 3 new: exception-message scrub, plain-message scrub, clean-log pass-through.

74/74 affected tests pass locally.

Not in this batch (deferred per reviewer):
- `_find_*` linear scan over `metadata_json` — fine at current scale, flagged for later.
- Full integration test exercising `_run_job_isolated → _execute_job → gate → ApprovalBlocked → blocked notif` end-to-end — the unit-level coverage now covers each leg, end-to-end can land in a follow-up if the dashboard's blocked rendering needs verification too.

backend/services/cron_scheduler.py
```diff
@@ -89,17 +108,23 @@ async def start(self):
...
+                # immediately re-evaluates the next due job. Same-job double-
+                # fires are prevented by `_inflight_tasks`.
+                job_id = next_job.id
+                if job_id in self._inflight_tasks and not self._inflight_tasks[job_id].done():
+                    logger.info(
+                        "[CRON] Job %s still in flight (likely awaiting approval) — "
+                        "skipping this fire", job_id,
+                    )
+                else:
+                    task = asyncio.create_task(self._run_job_isolated(job_id))
```
omnigentx, line 130: **[HIGH — regression introduced by this fix] Busy-spin / task starvation: overdue jobs will spin the loop at 100% CPU and never execute.**

The dispatch path has **no `await`** when `sleep_seconds == 0`, which is the steady state right after a job fires:

1. Job becomes due → the `wait_for` above times out → we reach here and `create_task(...)`. But `create_task` only *schedules* the coroutine — it runs only when the event loop regains control at an `await`.
2. `continue` → `_get_next_due()` (a **sync** method) returns the *same* job, because the task hasn't run yet so its `status` is still `"active"` and `next_run_at <= now` → `sleep_seconds = 0`.
3. `if sleep_seconds > 0:` is False → the `await wait_for(...)` is skipped → we fall into the `_inflight_tasks` branch → `logger.info("… skipping this fire")` → `continue`.
4. Back to step 2 forever. **No `await` is ever hit**, so the spawned `_run_job_isolated` task never gets scheduled → the job never flips to `running` → it's never removed from `_get_next_due()` → tight 100% CPU loop, log spam, and the cron *never actually runs*.

This is worse than the original blocking behaviour (which at least executed the job). It won't show up in the current unit tests because they call `_execute_job` directly rather than driving `start()`.

**Fix:** guarantee a yield each iteration and stop re-selecting the in-flight job. Minimum:
```python
task = asyncio.create_task(self._run_job_isolated(job_id))
self._inflight_tasks[job_id] = task
task.add_done_callback(lambda _t, jid=job_id: self._inflight_tasks.pop(jid, None))
await asyncio.sleep(0)   # yield so the task runs and flips status→running
```
and/or exclude `id in self._inflight_tasks` inside `_get_next_due()` so an in-flight job isn't returned again. Please add a test that drives `start()` for a couple of iterations (fake clock / wake event) so this path is covered.

omnigentx (commented):
## Re-review of the follow-up commits (`ab91755`, `992b18b`, `7676f06`)

Thanks for the fast turnaround. Verified each against source + ran the affected suites (`test_approval_gate` / `test_cron_scheduler_agent_turn` / `test_logging_filters` / `test_http_safety`) — **all pass**.

**Resolved correctly:**
- **#2 timeout persistence** — timeout now calls `resolve_approval(reject)`, so subsequent fires fast-skip via `_find_prior_decision` instead of re-blocking. ✅
- **#3 blocked status** — new `ApprovalBlocked` exception → run `status="blocked"`, distinct `blocked` notification, no `fail_count` bump / no auto-disable. Clean. ✅
- **#4 traceback redaction** — `RedactingFormatter` scrubs the fully rendered line (incl. `exc_info` traceback) on every handler; filter docstring updated. ✅
- **#5 uncapped fetches** — both `crawl_poller.py:201` and `:253` now go through `get_capped_text`. ✅

**One blocker remains — and the #1 fix introduces a new HIGH regression.** See the inline comment on `cron_scheduler.py`.

### Two more, lower priority

**[LOW] Startup catch-up still blocks.** `start()` still `await self._catch_up_missed()` (line ~80), and `_catch_up_missed` was not touched — it still runs `await self._execute_job(...)` serially on one shared session. A missed `agent_turn` job at boot can still freeze catch-up (and hold that session) for up to the gate timeout. Lower severity (startup only), but it's the same blocking pattern the loop fix was meant to remove — consider routing catch-up through `_run_job_isolated` too.

**[LOW / behavior] Timeout = permanent rejection.** Persisting timeout as `rejected` (good for #2) combined with "a rejected hash stays rejected" means a single missed approval window (e.g. nobody around for an hour) permanently blocks that cron payload — every later fire just emits a `blocked` notification with no re-prompt. Intended per the no-spam design, but it's a sharp edge; consider allowing re-approval, or a shorter gate timeout instead of a hard reject.

### Test-coverage note
The new unit tests exercise `_execute_agent_turn` / `_execute_job` directly (awaited), so they do **not** run through the real `start()` dispatch loop — which is exactly where the regression below lives. A test that drives `start()` for a few iterations with a fake clock/event would catch it.

omnigentx:
Addressed the re-review:

- **4e717c5** `fix(cron)` — closes the busy-spin regression (HIGH) and routes startup catch-up through `_run_job_isolated` (LOW):
  - `_get_next_due` now excludes ids in `_inflight_tasks` at the SQL level, so an overdue job whose task is queued-but-not-yet-running can't be re-selected on the next iteration.
  - `await asyncio.sleep(0)` after both branches of the dispatch (spawn AND skip), so the loop yields to the freshly spawned task before re-evaluating.
  - `_catch_up_missed` no longer awaits `_execute_job` serially on the shared session — it now spawns each missed run through `_run_job_isolated` with the same in-flight guard, matching the steady-state loop.
  - `gate()` docstring documents the timeout-as-permanent-reject trade-off (LOW design call) — keeping the no-spam behaviour explicit; clearing requires deleting the row or changing the payload.

Tests:
- `test_get_next_due_excludes_in_flight_jobs` — directly drives the regression (populate `_inflight_tasks["A"]`, mocked DB returns both A and B, assert next_due is B). Would fail on the previous commit's `_get_next_due`.
- `test_run_job_isolated_uses_fresh_db_session` — pins the "each dispatch gets its own session" invariant so a future change can't accidentally pin the loop's session for an hour.

5 tests in `test_cron_scheduler_agent_turn.py` pass; all 74 tests across the affected surfaces still pass.

omnigentx (commented):
## Round-3 re-review (`4e717c5`) — all blocking findings resolved ✅ LGTM

Verified against source + ran `test_cron_scheduler_agent_turn` + `test_approval_gate` (**17 passed**).

**HIGH busy-spin regression — fixed correctly (defense in depth):**
- `await asyncio.sleep(0)` added in *both* the dispatch and the in-flight-skip branches of `start()`, guaranteeing a yield every iteration. The spawned task now runs through `db.commit()` (status→`running`) before its first real `await`.
- `_get_next_due` now excludes `_inflight_tasks` ids via `~CronJobModel.id.in_(...)` (with an empty-list guard), so an overdue in-flight job is never re-selected.

Three independent layers (yield + `status="running"` filter + in-flight filter) — the tight loop is closed.

**Bonus: the LOW startup-catch-up finding is also fixed.** `_catch_up_missed` now collects ids, closes its session, then dispatches each through `_run_job_isolated` (own session) with the same in-flight skip + yield. No more serial blocking on one shared session at boot.

**Tests added** cover the root cause: `test_get_next_due_excludes_in_flight_jobs` and `test_run_job_isolated_uses_fresh_db_session`.

### Non-blocking, your call
- The exclusion + isolation are unit-tested, but there's still no test that drives the real `start()` loop for a few iterations and asserts bounded wake-ups. Nice-to-have, not required.
- **Timeout = permanent rejection** remains a product-level tradeoff (a single missed approval window hard-rejects that payload until it changes). Fine to keep as designed — just flagging it's a deliberate behavior, not an oversight.

Net: every blocking item from the previous two reviews (#01/#05 wiring + cron freeze + busy-spin, #02 timeout persistence, #03 blocked status, #04 traceback redaction, #06 fetch caps) is addressed. Good to merge from my side.

PR #58, opened 2026-06-01
feat(team-template): edit factory yaml + running team via UI + MCP

omnigentx:
## Summary

Adds two surfaces for editing team templates and exposes them to Jarvis via a dedicated MCP tool server. Per decisions on 2026-05-17 (yaml = factory default, not continuous source) and 2026-05-31 (UI lives in Settings; MCP uses the Unix-socket RPC bridge, no HTTP/API-key).

- **Factory yaml** (`team_templates/*.yaml`) — edit the recipe used at every team spawn.
- **Running team** (DB SSoT per `team_sessions`) — edit one team's template with audit, rollback, reset-to-yaml, and force-reload.
- **MCP server `team_template`** — 10 tools mirror the two surfaces; Jarvis can curate templates itself at runtime.

## Architecture

```
                                ┌─ HTTP (verify_api_key) ── Frontend UI
                                │
service layer ─── routes ──────┤
   (factory + running)          │
                                └─ RPC handlers ── UDS bridge ── MCP tools ── Jarvis LLM
```

- `services/team_template_factory_service.py` — yaml read/write/list with path-traversal guard, `yaml.safe_load` validation, `.bak` rotation, atomic rename (EBUSY fallback for Docker bind-mounts).
- `services/team_template_rpc_handlers.py` — 10 dotted methods (`team_template.factory.*`, `team_template.running.*`) registered on the existing `RuntimeRpcServer` next to skill/mcp/approval handlers.
- `tools/team_template_server.py` — FastMCP wraps each RPC method as a tool; registered as `mcp.servers.team_template` in `fastagent.config.yaml`.

## Design invariants

| Tool | Reads | Writes | Effect |
|---|---|---|---|
| `team_template_write_factory` | — | yaml file | Yaml only. Survives recreate. |
| `team_template_patch_role` | — | DB | DB only. Audited. Lost on recreate. |
| `team_template_reset_role` | yaml | DB | Pulls yaml → DB; yaml untouched. |
| `team_template_reload` | DB | — (process) | SIGKILL+respawn agents reading current DB. |

No tool writes both yaml and DB — Jarvis chains calls for combined updates. This is deliberate per decision 2026-05-17: factory yaml is the recipe, DB is the running team, and the two never auto-sync.

## Frontend

All under `Settings`:

- **YAML Config tab** — extended with a `TEAM TEMPLATES` group surfacing factory yamls. Editor + chrome reused; load/save branches by `kind` so `fastagent.config.yaml` flow is untouched.
- **Running Templates tab** (new) — session dropdown, drift pill, per-role editor (instruction / servers / skills / server_overrides / model / role_display), history rail with rollback. Reset-to-yaml opens a 3-way chooser (Cancel / Sync only / Sync + Force reload) so the user doesn't have to remember the two-step sequence. View-diff opens a `v-code-diff` modal with unified/side-by-side toggle (localStorage-persisted) and intra-line word highlight.

## Test plan

- [x] `tests/test_routes/test_team_templates_factory.py` (16 new) — auth, list, read (raw + parsed), write (atomic + .bak), malformed-yaml tolerance, path-traversal guard, oversize rejection
- [x] `tests/test_services/test_team_template_rpc_handlers.py` (15 new) — factory list/read/write, running get/patch/history/rollback/reset/yaml-diff/reload, registration count
- [x] `tests/test_routes/test_team_template_routes.py` + `tests/test_services/test_team_template_service.py` (48 pre-existing) — still pass
- [x] MCP server smoke test — `mcp.list_tools()` returns 10 expected tool names
- [x] Frontend `vite build` clean (SettingsView bundle: 100 → 237 KB; +40 KB gzip due to `v-code-diff` + highlight.js)
- [x] Backend restarted; `GET /api/team-templates` and `GET /api/team-sessions` return 401 (route exists) instead of 404 (route missing)

## Known UX gap

- No frontend Playwright e2e for the new tab yet. Manual click-flow (open Running Templates → pick session → edit role → save → audit row appears → rollback) verified locally.

### DIFF
--- a/backend/routes/team_template.py
+++ b/backend/routes/team_template.py
@@ -96,9 +96,15 @@ async def list_sessions():
     Used by the Settings → Running templates dropdown so the UI never has to
     pull the full template dict for every team just to render a selector.
     """
+    # Narrow the catch to ImportError so we degrade gracefully only when the
+    # fast_agent package is genuinely absent (e.g. a stripped test harness).
+    # Any runtime failure inside list_team_sessions() (DB down, corrupted
+    # row, etc.) must propagate as a 500 — silently returning an empty list
+    # would render the Settings dropdown blank with no signal that
+    # enumeration failed.
     try:
         from fast_agent.spawn.team_spawner import list_team_sessions
-    except Exception as exc:
+    except ImportError as exc:
         logger.warning("[team-template] team_spawner unavailable: %s", exc)
         return {"sessions": []}
     out = []
--- a/backend/routes/team_template.py
+++ b/backend/routes/team_template.py
@@ -86,6 +86,34 @@ class ResetBody(BaseModel):
 
 
 @router.get(
+    "",
+    dependencies=[Depends(verify_api_key)],
+)
+async def list_sessions():
+    """Lightweight enumeration of every team session.
+
+    Returns ``{"sessions": [{session_id, team_name, template_name, agents_count}]}``.
+    Used by the Settings → Running templates dropdown so the UI never has to
+    pull the full template dict for every team just to render a selector.
+    """
+    try:
+        from fast_agent.spawn.team_spawner import list_team_sessions
+    except Exception as exc:
+        logger.warning("[team-template] team_spawner unavailable: %s", exc)
+        return {"sessions": []}
+    out = []
+    for sess in list_team_sessions():
+        template = sess.get("template") or {}
+        out.append({
+            "session_id": sess.get("session_id"),
+            "team_name": sess.get("team_name"),
+            "template_name": template.get("name"),
+            "agents_count": len(sess.get("agents") or {}),
+        })
+    return {"sessions": out}
+
+
+@router.get(
     "/{session_id}/template",
     dependencies=[Depends(verify_api_key)],
 )
--- a/backend/services/team_template_factory_service.py
+++ b/backend/services/team_template_factory_service.py
@@ -1,0 +1,195 @@
+"""Factory-yaml team template service — read/write ``backend/team_templates/*.yaml``.
+
+Companion to :mod:`services.team_template_service`, which edits the running
+team's DB SSoT. This module edits the **factory default** the spawner reads
+at team creation. Per user decision 2026-05-17 the two paths are kept
+deliberately separate: editing a factory yaml does NOT mutate running teams.
+Drift is surfaced via ``/api/team-sessions/{id}/template/yaml-diff`` and the
+user clicks Reload-from-yaml manually.
+
+Safety properties (mirrors :mod:`routes.yaml_config`):
+
+* Only files directly inside the ``team_templates`` directory may be touched;
+  path-traversal (``..``) is refused.
+* Writes are ``yaml.safe_load``-validated before they hit disk so a broken
+  edit never bricks the next spawn.
+* Atomic write via ``tmp → os.replace`` with ``EBUSY`` fallback for bind-mounted
+  files (Docker compose dev loop).
+* Previous content rotates into a ``.bak`` sibling so a bad edit is recoverable.
+"""
+from __future__ import annotations
+
+import errno
+import logging
+import os
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+logger = logging.getLogger(__name__)
+
+# Resolved once at import — same hardening as yaml_config: even if cwd flips
+# mid-request we never escape the factory directory.
+_FACTORY_DIR = (Path(__file__).resolve().parent.parent / "team_templates").resolve()
+
+# Same cap as yaml_config.py; team templates are typically a few KB.
+MAX_BYTES = 256 * 1024
+
+
+class FactoryTemplateError(Exception):
+    """Base class for factory-yaml errors. Subclasses map to HTTP statuses."""
+
+    status: int = 500
+
+
+class NotFoundError(FactoryTemplateError):
+    status = 404
+
+
+class ValidationError(FactoryTemplateError):
+    status = 400
+
+
+class PathTraversalError(FactoryTemplateError):
+    status = 400
+
+
+def factory_dir() -> Path:
+    """Return the resolved team_templates directory. Mostly for tests."""
+    return _FACTORY_DIR
+
+
+def _resolve(name: str) -> Path:
+    """Map a logical template name (``agile_team``) to its yaml path.
+
+    Refuses any name that resolves outside ``_FACTORY_DIR`` — this is the
+    only allowlist we maintain. Adding a new template = dropping a yaml in
+    the directory; no service-side enum to keep in sync.
+    """
+    if not name or "/" in name or "\\" in name or name.startswith("."):
+        raise PathTraversalError(f"invalid template name: {name!r}")
+    candidate = (_FACTORY_DIR / f"{name}.yaml").resolve()
+    if _FACTORY_DIR not in candidate.parents:
+        raise PathTraversalError(f"path traversal blocked: {name!r}")
+    return candidate
+
+
+def list_factory_templates() -> list[dict[str, Any]]:
+    """Return every yaml in ``team_templates/`` with size + parsed display name.
+
+    Display name comes from the yaml's top-level ``name:`` field when it
+    exists (e.g. ``agile-team``). Falls back to the filename stem when the
+    file is empty or malformed so the UI can still show + offer to fix it.
+    """
+    if not _FACTORY_DIR.exists():
+        return []
+    out: list[dict[str, Any]] = []
+    for p in sorted(_FACTORY_DIR.glob("*.yaml")):
+        try:
+            text = p.read_text(encoding="utf-8")
+            doc = yaml.safe_load(text) or {}
+        except (OSError, yaml.YAMLError) as exc:
+            logger.warning("[factory-yaml] %s: %s", p.name, exc)
+            doc, text = {}, ""
+        team = doc.get("team") if isinstance(doc, dict) and isinstance(doc.get("team"), dict) else doc
+        display = (team.get("name") if isinstance(team, dict) else None) or p.stem
+        out.append({
+            "name": p.stem,
+            "filename": p.name,
+            "display_name": display,
+            "description": (team.get("description") if isinstance(team, dict) else None) or "",
+            "size": len(text.encode("utf-8")),
+            "exists": True,
+        })
+    return out
+
+
+def read_factory_template(name: str) -> dict[str, Any]:
+    """Return ``{name, filename, content, parsed, exists, size}``.
+
+    ``parsed`` is the safe-loaded dict (or ``None`` if file is empty or
+    invalid) so the UI can render a structural view without re-parsing.
+    """
+    path = _resolve(name)
+    if not path.exists():
+        raise NotFoundError(f"template not found: {name}")
+    try:
+        text = path.read_text(encoding="utf-8")
+    except OSError as exc:
+        raise FactoryTemplateError(f"read failed: {exc}") from exc
+    try:
+        parsed = yaml.safe_load(text)
+    except yaml.YAMLError as exc:
+        # Return content so the UI can still show it for hand-editing,
+        # but flag the parse failure so the UI can warn before save.
+        return {
+            "name": name,
+            "filename": path.name,
+            "content": text,
+            "parsed": None,
+            "parse_error": str(exc),
+            "exists": True,
+            "size": len(text.encode("utf-8")),
+        }
+    return {
+        "name": name,
+        "filename": path.name,
+        "content": text,
+        "parsed": parsed,
+        "exists": True,
+        "size": len(text.encode("utf-8")),
+    }
+
+
+def write_factory_template(name: str, content: str) -> dict[str, Any]:
+    """Validate + atomically write yaml. Rotates previous content to ``.bak``.
+
+    Returns ``{name, filename, size, saved: True}``. Raises:
+      * ValidationError — content fails ``yaml.safe_load`` or exceeds cap.
+      * PathTraversalError — name escapes the factory dir.
+      * FactoryTemplateError — underlying OSError.
+    """
+    if len(content.encode("utf-8")) > MAX_BYTES:
+        raise ValidationError(f"content exceeds {MAX_BYTES} bytes")
+
+    path = _resolve(name)
+    try:
+        yaml.safe_load(content)
+    except yaml.YAMLError as exc:
+        raise ValidationError(f"invalid YAML — file not saved: {exc}") from exc
+
+    if path.exists():
+        backup = path.with_suffix(path.suffix + ".bak")
+        try:
+            backup.write_bytes(path.read_bytes())
+        except OSError as exc:
+            logger.warning("[factory-yaml] could not write backup for %s: %s", path, exc)
+
+    tmp = path.with_suffix(path.suffix + ".tmp")
+    try:
+        tmp.write_text(content, encoding="utf-8")
+        try:
+            os.replace(tmp, path)
+        except OSError as exc:
+            # Docker bind-mounts a single file by inode — rename-over-mount
+            # fails with EBUSY on Linux. Fall back to truncate + in-place write;
+            # the .bak from above is still our rollback option.
+            if exc.errno != errno.EBUSY:
+                raise
+            path.write_text(content, encoding="utf-8")
+            try: tmp.unlink()
+            except OSError: pass
+    except OSError as exc:
+        if tmp.exists():
+            try: tmp.unlink()
+            except OSError: pass
+        raise FactoryTemplateError(f"write failed: {exc}") from exc
+
+    logger.info("[factory-yaml] %s saved (%d bytes)", path.name, len(content))
+    return {
+        "name": name,
+        "filename": path.name,
+        "size": len(content.encode("utf-8")),
+        "saved": True,
+    }
--- a/backend/services/team_template_factory_service.py
+++ b/backend/services/team_template_factory_service.py
@@ -11,11 +11,21 @@ Safety properties (mirrors :mod:`routes.yaml_config`):
 
 * Only files directly inside the ``team_templates`` directory may be touched;
   path-traversal (``..``) is refused.
-* Writes are ``yaml.safe_load``-validated before they hit disk so a broken
-  edit never bricks the next spawn.
+* Writes are validated for BOTH syntax (``yaml.safe_load``) and the minimal
+  template structure (top-level dict with a ``roles`` mapping, either at
+  the root or under ``team:``). The structural check is deliberately strict
+  because the MCP tool lets an unattended LLM write these — a syntactically
+  valid but structurally wrong yaml would brick every subsequent
+  ``spawn_team`` call.
 * Atomic write via ``tmp → os.replace`` with ``EBUSY`` fallback for bind-mounted
   files (Docker compose dev loop).
 * Previous content rotates into a ``.bak`` sibling so a bad edit is recoverable.
+
+Concurrency: writes are single-writer (no file lock). The MCP tool plus the
+UI can in principle race on the same file, but the only realistic concurrent
+caller pair is "human in the UI" + "human-driven LLM", which is implicitly
+serialised by the human. If true concurrent automated writers ever land,
+gate this module behind a per-path lock.
 """
 from __future__ import annotations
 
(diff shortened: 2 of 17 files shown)


omnigentx:
## Code Review — Team template edit (factory yaml + running team + MCP)

**Verdict:** Strong, well-documented PR. The factory-yaml vs. running-team-DB SSoT separation is rigorously enforced and the wiring follows existing patterns exactly. A few correctness/doc mismatches and one project-policy gap (e2e) to address before merge.

### What's done well
- **SSoT discipline:** the "no tool writes both yaml and DB" invariant is real and enforced — `reset_role`/`reload` are the only yaml→DB paths, both user-gated. Code matches the design table in the description.
- **Path-traversal guard** (`team_template_factory_service.py:63-75`): rejects `/`, `\`, and `.`-prefixed names, then `.resolve()` + parent-check defeats symlink escapes. `verify_api_key` on all three REST routes.
- **Async RPC:** `_running_reload` is `async`; `runtime_rpc._dispatch_line` handles coroutine handlers (`iscoroutinefunction`), so dispatch is correct.
- **History limit** clamped at the DB layer (`team_template_service.py:333`, `max(1, min(limit, 500))`) — MCP docstring "1..500" holds.
- **Frontend:** friendly error surfacing throughout, 3-way reset chooser, confirm-gated destructive reload. The JSON `catch` at `SettingsRunningTemplates.vue:183` re-throws a friendly message — not a silent swallow.

### Issues / risks

**1. Docstring overstates the write guarantee (medium — correctness/doc)**
`write_factory_template` validates only YAML *syntax* (`yaml.safe_load`, result discarded), but the module docstring (`team_template_factory_service.py:14-15`) claims *"a broken edit never bricks the next spawn."* A syntactically-valid but structurally-wrong yaml (e.g. missing `team.roles`) saves fine and **will** break the next spawn. Either soften the docstring to "syntactically valid" or add minimal structural validation (top-level `team`/`roles` present).

**2. No e2e test for the new UI tab (medium — project policy)**
"Happy Path FIRST" makes ≥1 real click-flow e2e a hard rule for user-facing changes. The 1214-line `SettingsRunningTemplates.vue` (session-pick → edit → save → audit → rollback → reset/reload) ships with zero Playwright coverage. The manual-verification note is honest, but this is the one item that should block per the project guidelines. At minimum add the happy-path spec or get explicit sign-off to defer.

**3. Clearing `server_overrides` in the UI is a no-op (minor — UX)**
`SettingsRunningTemplates.vue:179` — `_buildPatch` only considers `server_overrides` when `server_overrides_text.trim()` is truthy. If a user empties the field to *remove* overrides, no patch field is sent, so the override silently persists. Consider: empty text + non-empty `live.server_overrides` → send `{}`.

**4. Silent empty-list on spawner import failure (minor — silent fallback)**
`routes/team_template.py` `list_sessions` returns `{"sessions": []}` on `ImportError` with only a `logger.warning`. The dropdown then renders empty with no UI signal that enumeration failed vs. genuinely-no-sessions. Acceptable as a defensive import guard, but worth a comment clarifying this is import-failure-only (runtime errors in `list_team_sessions()` correctly propagate as 500).

**5. No write concurrency guard (low)**
`write_factory_template` has no lock — two concurrent writes could interleave `.bak`/`.tmp`. Low risk for a single admin, but the MCP tool means Jarvis *and* the UI can both write. Mention-only.

### Test coverage
- Backend: 16 (factory routes) + 15 (RPC handlers) new, +48 pre-existing — good breadth (auth, traversal, malformed-yaml, atomic+bak, oversize, registration count).
- **Gap:** frontend e2e (see #2). Also confirm explicit cases exist for the `_factory_yaml_path_for` hyphen↔underscore mapping (`agile-team` → `agile_team.yaml`), a quiet correctness dependency in reset/diff.

### Recommendation
Address **#1** (one-line docstring fix or small validation) and resolve **#2** (add the e2e or document the deferral with sign-off). #3–#5 are non-blocking. Architecture and security are sound — mergeable once the e2e policy gap is closed.

backend/services/team_template_factory_service.py
```diff
@@ -0,0 +1,195 @@
...
+      * ValidationError — content fails ``yaml.safe_load`` or exceeds cap.
+      * PathTraversalError — name escapes the factory dir.
+      * FactoryTemplateError — underlying OSError.
+    """
+    if len(content.encode("utf-8")) > MAX_BYTES:
+        raise ValidationError(f"content exceeds {MAX_BYTES} bytes")
+
+    path = _resolve(name)
+    try:
+        yaml.safe_load(content)
```
omnigentx: **[BLOCKER — correctness] Validate structure, not just syntax.**

This only checks the content *parses* as YAML; the result is discarded. A syntactically-valid but structurally-wrong template (e.g. missing `team.roles`, wrong types) saves fine and **bricks the next team spawn**. The module docstring (L14-15) claims "a broken edit never bricks the next spawn" — that guarantee does not hold today.

Because `team_template_write_factory` lets the LLM write this unattended, please make the guard strict:
- after `safe_load`, assert the doc is a dict and the expected top-level shape exists (`team`/`roles`), raising `ValidationError` otherwise;
- then either soften the docstring to "syntactically valid" *or* (preferred) keep the strong claim and back it with the structural check.

backend/services/team_template_factory_service.py
```diff
@@ -0,0 +1,195 @@
...
+    except yaml.YAMLError as exc:
+        raise ValidationError(f"invalid YAML — file not saved: {exc}") from exc
+
+    if path.exists():
+        backup = path.with_suffix(path.suffix + ".bak")
```
omnigentx, line 220: **[low — concurrency]** No lock around the `.bak` rotation + `.tmp` + replace sequence. Single admin is fine, but the MCP tool means the LLM *and* the UI can both write concurrently, which could interleave the backup/tmp files. Worth a short comment documenting the single-writer assumption, or a simple per-path lock if concurrent writes are realistic.

backend/routes/team_template.py
```diff
@@ -85,6 +85,34 @@ class ResetBody(BaseModel):
...
+
+    Returns ``{"sessions": [{session_id, team_name, template_name, agents_count}]}``.
+    Used by the Settings → Running templates dropdown so the UI never has to
+    pull the full template dict for every team just to render a selector.
+    """
+    try:
+        from fast_agent.spawn.team_spawner import list_team_sessions
+    except Exception as exc:
+        logger.warning("[team-template] team_spawner unavailable: %s", exc)
+        return {"sessions": []}
```
omnigentx, line 109: **[minor — silent fallback]** On `ImportError` this returns an empty session list with only a `logger.warning`, so the dropdown renders empty with no UI signal that *enumeration failed* vs. *genuinely no sessions*. Please narrow the catch to `ImportError` (runtime errors inside `list_team_sessions()` should still surface as 500, not be swallowed), and add a comment that this branch is import-failure-only.

frontend/src/views/settings/SettingsRunningTemplates.vue
```diff
@@ -0,0 +1,1214 @@
...
+  const liveServers = [...(live.servers || [])].sort()
+  if (JSON.stringify([...servers].sort()) !== JSON.stringify(liveServers)) {
+    patch.servers = servers
+  }
+  const skills = _splitLines(draft.value.skills_text)
+  const liveSkills = [...(live.skills || [])].sort()
+  if (JSON.stringify([...skills].sort()) !== JSON.stringify(liveSkills)) {
+    patch.skills = skills
+  }
+  if (draft.value.server_overrides_text.trim()) {
```
omnigentx: **[minor — UX correctness]** `_buildPatch` only considers `server_overrides` when the textarea is non-empty. If a user clears the field to *remove* overrides, no patch field is sent and the existing override silently persists. Handle the clear case: empty text + non-empty `live.server_overrides` → send `server_overrides: {}` (matching how `servers`/`skills` already send empty arrays).

omnigentx (commented):
## ⚠️ Changes requested — tighten before merge

**Coder note (please work rigorously here):** this PR exposes write paths to *both* the UI and the Jarvis LLM (via MCP), so a weak guard becomes a dangerous bug, not a cosmetic one. Two classes of danger to close hard:

1. **A bad template can brick every future team spawn.** The factory-yaml write currently validates only YAML *syntax*, not *structure*. A syntactically-valid file missing `team.roles` saves cleanly and breaks the next spawn — and the LLM can trigger this unattended via `team_template_write_factory`. Add minimal structural validation (assert top-level `team`/`roles` present) and correct the docstring claim. Treat "valid input" as "valid template", not "parses as YAML".
2. **No real e2e on a 1200-line user-facing tab.** Per the project's "Happy Path FIRST" rule this should block. The destructive `reload` (SIGKILL+respawn) is reachable from this UI — a basic broken click-flow here kills running agents. Need ≥1 Playwright spec on the real flow (pick session → edit → save → audit → rollback → reset/reload).

Non-blocking items (#3–#5) are inline. Architecture and the SSoT separation are sound — these are about making the guards strict enough that neither a fat-fingered admin nor the LLM can cause damage.

Full write-up posted as a separate comment above.

backend/services/team_template_factory_service.py
omnigentx: Fixed in 5efa46e. `_validate_template_structure` now runs after `safe_load` and rejects: non-mapping top-level, missing/non-mapping `roles`, empty roles, list-shaped roles, non-mapping role configs. Docstring updated to reflect the strong claim. Six new unit tests cover each rejection path + the two layouts the spawner accepts.

backend/services/team_template_factory_service.py
omnigentx: Acknowledged. Documented the single-writer assumption in the module docstring (5efa46e) — UI + LLM are realistic concurrent callers in theory but are both driven by the same human in practice. Will revisit with a per-path lock if true automated concurrent writers ever land.

backend/routes/team_template.py
omnigentx: Fixed in 5efa46e. Narrowed to `except ImportError` with an inline comment explaining the import-failure-only intent. Runtime errors inside `list_team_sessions()` now propagate as 500 instead of being silently swallowed.

frontend/src/views/settings/SettingsRunningTemplates.vue
omnigentx: Fixed in 5efa46e. Empty/whitespace `server_overrides_text` now round-trips to `{}` in both `_buildPatch` and the `dirty` computed — user can actually clear overrides via the UI, and the UNSAVED pill stays off when the change is a true no-op.

omnigentx:
Review fixes pushed in 5efa46e. Summary:

| # | Severity | Status |
|---|---|---|
| 1 | BLOCKER — structural validation | ✅ Fixed: `_validate_template_structure` rejects non-mapping top-level, missing `roles`, list-shaped `roles`, non-mapping role configs, empty roles. +6 unit tests covering rejection paths + both spawner layouts. Docstring updated. |
| 2 | low — concurrency | ✅ Documented: single-writer assumption explicit in module docstring; will revisit with a per-path lock if true automated concurrent writers ever land. |
| 3 | minor — ImportError silent fallback | ✅ Fixed: narrowed to `except ImportError`; runtime errors propagate as 500 now. |
| 4 | minor — clear server_overrides | ✅ Fixed: empty/whitespace round-trips to `{}` in both `_buildPatch` and the `dirty` computed. |
| policy | Happy-Path-FIRST e2e | ✅ Added: `flows/settings-running-templates.spec.ts` — pick session → edit role → save → audit row → rollback. 5.5s, passes. |

Test counts:
- Backend: 85 (was 79; +6 structural-validation cases)
- Frontend e2e: 1 new spec passes
- Frontend build: clean

All 4 inline threads resolved.

backend/services/team_template_rpc_handlers.py
```diff
@@ -0,0 +1,303 @@
...
+    return SessionLocal()
+
+
+def _factory_yaml_path_for(name: str) -> Path:
+    """Resolve the yaml path the running template was instantiated from.
+
+    Mirrors ``_resolve_yaml_for_session`` in ``routes/team_template.py``: maps
+    e.g. ``agile-team`` → ``agile_team.yaml``. Caller is the reset/diff handler.
+    """
+    candidate = factory_svc.factory_dir() / f"{(name or 'agile-team').replace('-', '_')}.yaml"
```
omnigentx, line 53: **[normal — security / path traversal] `_factory_yaml_path_for` bypasses the `_resolve()` guard.**

This builds the yaml path by raw string interpolation (`factory_dir() / f'{name.replace("-","_")}.yaml'`), skipping `factory_svc._resolve()` which rejects `/`, `\`, leading-`.` names and verifies the resolved path stays inside the factory dir.

`name` flows from `template['name']` (DB), populated at spawn time from the yaml's top-level `name:` field — and `_validate_template_structure` only checks `roles:`, **never `name:`**. Since `team_template_write_factory` lets the LLM write factory yaml unattended, a `name: ../../target` value makes `reset_role`/`yaml_diff` `open()` files **outside** `team_templates/`.

**Fix:** route both candidates through `factory_svc._resolve(name.replace('-', '_'))` (catch `PathTraversalError`). The same flawed pattern pre-exists in `routes/team_template.py::_resolve_yaml_for_session` — worth fixing in lockstep.

frontend/src/views/settings/SettingsRunningTemplates.vue
```diff
@@ -0,0 +1,1234 @@
...
+  successMsg.value = ''
+  try {
+    const r = await _doResetCall()
+    let respawned = 0
+    if (mode === 'sync-and-reload') {
+      reloading.value = true
+      try {
+        const rl = await _doReloadCall()
+        respawned = Object.values(rl.results || {}).flat().length
+      } finally {
```
omnigentx, line 290: **[normal — data loss] sync-and-reload partial failure leaves UI stale → silent overwrite.**

If reset (step 1) succeeds but `_doReloadCall()` (step 2) throws, the inner `try/finally` only clears the spinner — the exception bubbles to the outer `catch`, which sets an error but **never refetches**. The DB now holds the yaml-synced template, but `template`/`draft`/`history` still show the pre-reset state.

User reads it as 'nothing happened', edits the stale fields and clicks Save → `_buildPatch` diffs against the stale `live` and PATCHes over the just-synced DB state, silently undoing the sync.

**Fix:** call `loadTemplate(activeSessionId.value)` in a `finally` (or in the catch) so the UI always converges to DB truth, and tailor the message to distinguish 'sync applied but reload failed' from 'sync failed'.

frontend/src/views/settings/SettingsRunningTemplates.vue
```diff
@@ -0,0 +1,1234 @@
...
+function jsonPreview(v) {
+  const s = JSON.stringify(v)
+  return s.length > 60 ? `${s.slice(0, 57)}…` : s
+}
+
+
+// ── Lifecycle ────────────────────────────────────────────────────────────
+
+watch(activeSessionId, (sid) => sid && loadTemplate(sid))
+watch(activeRole, () => _seedDraft())
```
omnigentx, line 446: **[normal — data loss] Switching role/session/Refresh discards unsaved edits with no confirm.**

`watch(activeRole, () => _seedDraft())` and `watch(activeSessionId, ... loadTemplate ...)` re-seed `draft` from the live template with no dirty-check; the role list uses `@click="activeRole = role"` directly. Clicking another role mid-edit throws away the typed text and the UNSAVED pill just vanishes.

**Fix:** mirror the existing `useConfirm()` guard in `SettingsYaml.vue::selectFile` — wrap the role-click handler, the session-dropdown change, and the Refresh button (or guard inside `_seedDraft` when `dirty`).

backend/services/team_template_rpc_handlers.py
```diff
@@ -0,0 +1,303 @@
...
+    except svc.NotFoundError as exc:
+        return _err(str(exc), 404)
+
+    yaml_path = _factory_yaml_path_for(current_template.get("name") or "agile-team")
+    if not yaml_path.exists():
+        return _err(f"factory yaml not found at {yaml_path}", 404)
+
+    with yaml_path.open(encoding="utf-8") as f:
+        yaml_doc = yaml_lib.safe_load(f) or {}
+    yaml_template = yaml_doc.get("team") or yaml_doc
```
omnigentx, line 223: **[nit — robustness] `yaml_diff` raises `AttributeError`→500 on a non-dict `team:`.**

`yaml_template = yaml_doc.get("team") or yaml_doc` has no `isinstance` guard, so a truthy non-dict `team:` (e.g. `team: [pm, dev]`) makes the next `.get("roles")` raise. The sibling `list_factory_templates` (factory_service.py:105-106) already guards this with `isinstance(...)`.

**Fix:** `yaml_template = yaml_doc.get("team") if isinstance(yaml_doc.get("team"), dict) else yaml_doc`, then bail with a 400 if it still isn't a dict.

frontend/src/views/settings/SettingsRunningTemplates.vue
```diff
@@ -0,0 +1,1234 @@
...
+    )
+  } catch {
+    // Unparseable text is definitively dirty — Save will surface the error.
+    draftOverrides = draft.value.server_overrides_text
+  }
+  return (
+    draft.value.instruction !== (live.instruction || '') ||
+    draft.value.role_display !== (live.role_display || '') ||
+    draft.value.model !== (live.model || '') ||
+    draft.value.servers_text !== (live.servers || []).join('\n') ||
```
omnigentx, line 90: **[nit — UX] Reordering servers/skills shows UNSAVED but Save says 'No changes' → stuck loop.**

The `dirty` check compares `join('\n')` (order-sensitive) while `_buildPatch` sorts both sides (order-insensitive, matching the backend's set semantics in `compute_role_diff`). A pure reorder is dirty-but-empty-patch: pill stays lit, Save no-ops, no refetch.

**Fix:** make the `dirty` check sort/normalize servers & skills the same way `_buildPatch` does.

frontend/src/views/settings/SettingsRunningTemplates.vue
```diff
@@ -0,0 +1,1234 @@
...
+          v-if="resetModal.visible"
+          class="reset-overlay"
+          @click.self="onResetModalCancel"
+        >
+          <div class="reset-card">
+            <div class="reset-icon">↺</div>
+            <h3 class="reset-title">Reset <code>{{ activeRole }}</code> to factory yaml?</h3>
+            <p class="reset-body">
+              Pulls every editable field of <strong>{{ activeRole }}</strong> from
+              <code>team_templates/{{ template?.name?.replace('-', '_') || 'agile_team' }}.yaml</code>
```
omnigentx, line 707: **[nit — cosmetic] `.replace('-', '_')` only replaces the first hyphen.**

JS `String.prototype.replace` with a string pattern is single-shot, but the backend uses Python `str.replace` (replaces all). For `my-cool-team` the modal shows `my_cool-team.yaml` while the backend resolves `my_cool_team.yaml`. Display-only (reset API computes the path server-side).

**Fix:** `.replaceAll('-', '_')` or `.replace(/-/g, '_')`.

frontend/src/views/settings/SettingsYaml.vue
```diff
@@ -202,11 +268,14 @@ onBeforeUnmount(() => window.removeEventListener('keydown', onKeydown))
...
-          :key="f.name"
+          :key="`${f.kind}:${f.name}`"
           type="button"
           class="file-item"
-          :class="{ active: activeName === f.name, dirty: activeName === f.name && dirty }"
-          @click="selectFile(f.name)"
+          :class="{
+            active: activeName === f.name && activeKind === f.kind,
+            dirty: activeName === f.name && activeKind === f.kind && dirty,
+          }"
```
omnigentx, line 277: **[nit — cosmetic] Dirty-dot `●` ignores `kind`, cross-triggers on same-named files.** *(anchored here on the `:class` change; the actual bug is the `dirty-dot` span ~L286, just past the diff hunk.)*

This `:class` correctly scopes `active`/`dirty` to `name` **AND** `kind`, but the inline `<span v-if="activeName === f.name && dirty" class="dirty-dot">●</span>` (~L286) still checks only `name`. Since this PR merges fastagent + team-template files into one tree, a team template named `config`/`secrets` (allowed by `_resolve`) collides and lights the `●` on both rows. Active-border + Save logic stay correctly kind-scoped, so this is display-only.

**Fix:** add `activeKind === f.kind` to that span's `v-if` to mirror this `:class`.

omnigentx:
## Ultrareview — 7 findings (verified against `5efa46e`)

All 7 confirmed by reading the actual code on the latest commit, not the commit message. Inline comments are attached to each line. Priority:

### 🔴 Fix before merge (normal — the dangerous class)
1. **Path traversal via unvalidated `name:`** — `team_template_rpc_handlers.py::_factory_yaml_path_for` builds the yaml path by raw string interpolation, bypassing `factory_svc._resolve()`. `name` comes from the yaml's top-level `name:`, which `_validate_template_structure` never checks. An LLM (MCP `team_template_write_factory`, no human gate) can write `name: ../../target` → `reset_role`/`yaml_diff` `open()` files outside `team_templates/`. *(Same pattern pre-exists in `routes/team_template.py::_resolve_yaml_for_session` — fix in lockstep.)*
2. **sync+reload partial failure → silent overwrite** — `SettingsRunningTemplates.vue`: if reset succeeds but reload throws, the catch never refetches; UI shows stale state, user re-edits and Saves over the just-synced DB. Fix: refetch in `finally` + tailored "sync applied but reload failed" message.
3. **Switching role/session discards unsaved edits, no confirm** — `SettingsRunningTemplates.vue` watchers re-seed `draft` with no dirty-guard; `SettingsYaml.vue::selectFile` already has the `useConfirm()` pattern to copy.

### 🟡 Nit (fix-with or follow-up)
4. `yaml_diff` `AttributeError`→500 on non-dict `team:` — missing `isinstance` guard (sibling `list_factory_templates` already guards it).
5. Reorder servers/skills: `dirty` order-sensitive vs `_buildPatch` order-insensitive → UNSAVED pill stuck, Save says "No changes".
6. `.replace('-','_')` replaces only the first hyphen (JS) vs all (Python) → modal shows wrong filename. Display-only.
7. Dirty-dot `●` checks only `name`, not `kind` → cross-triggers on same-named files across the merged tree. Display-only.

**Recommendation:** close #1–#3 before merge (1 security, 2 data-loss; all small fixes). #4–#7 can be batched in or tracked as follow-up. Architecture/SSoT separation remain sound.

PR #59, opened 2026-06-01
fix: story audio playback, crawl redesign, voice STT, interrupt, passkey login

omnigentx:
Bundles the bug fixes + crawl rework from this session. Each is its own commit; all verified on top of latest `main` (backend 51 tests pass, frontend build + 77 unit tests pass).

## Commits
1. **feat(crawl): site-agnostic AI-driven story crawl + serpapi + dead-code cleanup**
   Rewrite crawling to work on ANY site/language: `detect_story_structure` renders JS + hands the LLM a language-agnostic structural summary → a small "recipe"; the worker enumerates the full chapter list from it (1000s of URLs never touch the LLM → fixed token cost). LIST + CHAIN modes with a same-story guard (no drift into other stories). `verify_chapters` checks the first 3 chapters. Self-heal: a stuck crawl wakes CrawlStoriesAgent to fix the recipe + `resume_crawl` from checkpoint. Added serpapi to the agent (fast search). Removed ~500 lines of dead crawl code. Verified e2e on truyenfull.today (LIST/VN), truyencom.com (CHAIN/VN/JS), webnovel.com (EN/JS). Preserves main's `get_capped_text` security fix.
2. **fix(audio): unify story playback on the singleton player + stop ~11s false-end**
   Chat/story audio routes through the one audioPlayer (persists across nav, mini-player). Backend no longer cancels the chapter being listened to; player treats a premature `ended` during live TTS as an underrun (resume) not end-of-chapter. Capped network retries.
3. **fix(chat/voice): one story-playback path, no leaked tags, no lost conversations**
   Shared `resolve_story_playback` for typed + voice; strip all backend marker tags before the bubble (except `[[[PLAY]]]`); one canonical READ_LOCAL tag; structured `crawl_job_id`; fix "No Agent" on create/delete; cancelled first turn still stamps `primary_agent` (was hiding the conversation); re-fetch on auth RESTORED.
4. **fix(interrupt): Stop while delegating a sub-agent no longer dead-locks chat**
   Track the whole-turn task (`_turn_tasks`) so interrupt/hard_kill cancel even mid-delegate, instead of leaking `_agent_lock` and freezing all later chats.
5. **fix(stt): Soniox keepalive + reconnect so voice doesn't silently die**
   Send keepalive every 10s during silence; classify fatal (400/401/403) vs transient (408 etc → reconnect). Layered on top of main's STT lifecycle/state-machine.
6. **fix(auth): passkey login no longer 403s on a stale CSRF cookie**
   Exempt `/api/auth/passkey/authenticate` from CSRF (login flow, no token yet); REGISTER stays protected.
7. **fix(ui): AuthGate spacing + chapter play/pause button**
   Wordmark no longer overlaps the login card; chapter button pauses/resumes the current chapter instead of restarting.

## Verification
- backend: `pytest` session/soniox/tag-relay/audio-reader → 51 passed
- frontend: `npm run build` ✓, `npm run test:unit` → 77 passed / 0 failed
- crawl e2e on 3 real sites (LIST, CHAIN, EN/JS)

### DIFF
--- a/backend/core/csrf.py
+++ b/backend/core/csrf.py
@@ -46,10 +46,13 @@ Exemptions
 ----------
 
 * ``GET / HEAD / OPTIONS`` are always allowed.
-* The auth bootstrap endpoints (``/api/auth/login``, ``/api/setup/...``)
-  are exempt because the user has not yet been issued a CSRF cookie at
-  that point.  ``/api/auth/login`` itself is rate-limited per-IP and
-  same-origin enforced by SameSite=Lax on the response Set-Cookie.
+* The auth bootstrap endpoints (``/api/auth/login``,
+  ``/api/auth/passkey/authenticate/*``, ``/api/setup/...``) are exempt because
+  the user has not yet been issued a (valid) CSRF cookie at that point — a
+  stale cookie from an expired session must not block re-login. Passkey
+  authenticate is defended by WebAuthn's own challenge/signature; ``login`` is
+  rate-limited per-IP and same-origin enforced by SameSite=Lax. Passkey
+  REGISTER stays protected (needs an authenticated session).
 * OAuth callback endpoints (``/api/oauth/.../callback``) are exempt —
   third-party redirect cannot carry our header.  Those routes have
   their own ``state``-parameter CSRF defence.
@@ -75,6 +78,15 @@ _PROTECTED_METHODS: frozenset[str] = frozenset({"POST", "PUT", "PATCH", "DELETE"
 _EXEMPT_PREFIXES: tuple[str, ...] = (
     "/api/auth/login",       # user has no CSRF cookie yet
     "/api/auth/logout",      # safe to invoke without prior cookie (no side effect on logout)
+    # Passkey AUTHENTICATE is a LOGIN flow (same class as /api/auth/login): the
+    # user is not signed in yet, so they have no valid CSRF token to send in the
+    # header. A stale CSRF cookie left over from an expired session would
+    # otherwise make this 403 ("has cookie, header empty") and lock the user out
+    # of re-logging-in. WebAuthn's own challenge/signature is the CSRF defence
+    # here. NOTE: passkey REGISTER is intentionally NOT exempt — it requires an
+    # authenticated session + real CSRF token (don't let an attacker silently
+    # enrol a passkey).
+    "/api/auth/passkey/authenticate",
     "/api/setup",            # bootstrap surface; user has no CSRF cookie until wizard ends
     "/api/oauth/google/callback",  # third-party redirect; has its own `state` CSRF
 )
--- a/backend/routes/chat.py
+++ b/backend/routes/chat.py
@@ -13,7 +13,7 @@ from pydantic import BaseModel
 
 from core.auth import verify_api_key
 from helpers.text_processing import process_agent_response, clean_text_for_tts
-from helpers.story_reader import handle_read_local, check_pending_read
+from helpers.story_reader import handle_read_local, check_pending_read, build_story_meta
 from services.shared_state import (
     session_service, library_manager, tts_cache,
 )
@@ -111,11 +111,21 @@ class ChatResponse(BaseModel):
     playback_url: str | None = None
     user_text: str | None = None
     conversation_id: str | None = None
+    # Present when the reply is a local story chapter, so the frontend can
+    # route playback through the singleton story player (chapter nav, resume)
+    # instead of treating it as a one-off TTS clip.
+    story: dict | None = None
+    # Present when the reply started a web crawl, so the frontend can poll
+    # /api/stories/crawl/status/{id} for live progress + a cancel button.
+    crawl_job_id: str | None = None
 
 
-def _process_response_tags(response_text, tts_text, book_id):
-    """Process story/library/local tags in response text. Returns (response_text, tts_text, book_id)."""
-    
+def _process_response_tags(response_text, tts_text, book_id, story_meta=None):
+    """Process story/library/local tags in response text.
+
+    Returns ``(response_text, tts_text, book_id, story_meta)``.
+    """
+
     # Check READ_STORY tag
     if not book_id:
         match = re.search(r"\[\[\[READ_STORY: (.*?)\]\]\]", response_text)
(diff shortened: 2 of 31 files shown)


backend/services/pause_controller.py
```diff
@@ -503,7 +511,12 @@ async def on_before_llm_call(runner: Any, messages: Any) -> None:
             # than wrapping the LLM call) because hooks are already in
             # the call site and we don't have to touch tool_runner.
             try:
-                self._current_tasks[agent_name] = asyncio.current_task()
+                cur = asyncio.current_task()
+                self._current_tasks[agent_name] = cur
+                # Track the turn-level task too (first LLM call of the turn).
+                # interrupt() uses this so a hard Stop works during the tool/
+                # delegate phase, when _current_tasks is empty.
+                self._turn_tasks.setdefault(agent_name, cur)
```
omnigentx, line 519: **[normal — verify with a test] `_turn_tasks.setdefault` assumes one task per turn.**

This is correct **only if** every `before_llm_call` within a turn fires in the *same* asyncio task (turn loop = one coroutine, hooks inline). If a later LLM call ran in a different task, `setdefault` keeps the first (possibly already-done) task → `interrupt()`/`hard_kill()` cancel a done task → the delegate-deadlock regresses silently.

The fast-agent architecture makes this hold today, but the deadlock this fixes was severe (held `_agent_lock`, froze all later chats). **Please add an integration test that interrupts a turn *mid-delegate*** (not mid-LLM) and asserts the lock is released — that's the one path with no coverage right now.

backend/core/csrf.py
```diff
@@ -75,6 +78,15 @@
...
     "/api/auth/logout",      # safe to invoke without prior cookie (no side effect on logout)
+    # Passkey AUTHENTICATE is a LOGIN flow (same class as /api/auth/login): the
+    # user is not signed in yet, so they have no valid CSRF token to send in the
+    # header. A stale CSRF cookie left over from an expired session would
+    # otherwise make this 403 ("has cookie, header empty") and lock the user out
+    # of re-logging-in. WebAuthn's own challenge/signature is the CSRF defence
+    # here. NOTE: passkey REGISTER is intentionally NOT exempt — it requires an
+    # authenticated session + real CSRF token (don't let an attacker silently
+    # enrol a passkey).
+    "/api/auth/passkey/authenticate",
```
omnigentx, line 89: **[test gap] Passkey-authenticate CSRF exemption — solid reasoning, no test.**

The exemption is justified (login flow, no token yet; WebAuthn challenge/signature is the defence; REGISTER stays protected). But there's no test pinning either half of the invariant. **Please add:** (1) `POST /api/auth/passkey/authenticate` succeeds with a stale/missing CSRF cookie; (2) `POST /api/auth/passkey/register` still 403s without a valid token. This is the cheapest of the three test gaps and guards a security boundary against a future refactor flipping it.

backend/services/stt_backends/soniox.py
```diff
@@ -83,13 +92,31 @@ def _on_connected(self) -> None:
...
-                "detail": data.get("error_message") or f"soniox error {err_code}",
-            })
+            # Transient errors (e.g. 408 idle timeout) are recoverable — the
+            # base loop reconnects. Only bad-key/bad-config codes are fatal.
+            try:
+                fatal = int(err_code) in _SONIOX_FATAL_ERROR_CODES
+            except (TypeError, ValueError):
+                fatal = False
+            if fatal:
+                self.mark_fatal()
```
omnigentx, line 111: **[test gap] fatal-vs-transient classification is uncovered.**

The 400/401/403→`mark_fatal()` (stop reconnect) vs everything-else→reconnect split is the crux of the 'voice silently dies' fix, plus the 10s keepalive in the base. No test exercises either branch. **Please add** a unit test feeding a fake 403 event (asserts `_fatal` set, no user-facing error suppression bug) and a 408 event (asserts reconnect, no `error` emit).

backend/routes/tts.py
```diff
@@ -139,15 +142,32 @@ async def tts_endpoint(request_id: str, request: Request, _auth=Depends(verify_o
...
-        
+
         headers = {"Accept-Ranges": "bytes", "Content-Type": "audio/mpeg"}
         if file_size > 0:
             headers["Content-Length"] = str(file_size)
+        # Tell the player whether this audio is still being generated. The
+        # frontend uses this on a premature 'ended' to tell a live-stream
+        # buffer underrun (resume) apart from a real end-of-chapter (advance).
+        if os.path.exists(cache_path + ".lock"):
+            headers["X-TTS-Generating"] = "1"
```
omnigentx, line 153: **[praise / SSoT] Good root-cause fix.**

The old `is_generating(request_id)` never matched (pre-gen keys by `{story_title, chapter_file}`, not `request_id`) so it *always* fell through to `request_cancel()`, killing the very chapter the user requested → the ~11s false-end. Switching to the `.lock` on this `cache_path` as the single source of truth is the right fix, and this `X-TTS-Generating` header cleanly pairs with the frontend HEAD-probe. No change requested — flagging the SSoT contract so a future reader doesn't reintroduce the request_id check.

backend/routes/chat.py
```diff
@@ -180,12 +190,85 @@ def _process_response_tags(response_text, tts_text, book_id):
...
 
-    return response_text, tts_text, book_id
+    # Final sweep: strip EVERY backend marker tag so none reach the chat bubble
+    # as raw text. The per-tag branches above only strip on a successful match
+    # AND when book_id wasn't already set — but find_story_chapter sets book_id
+    # via the pending-read queue FIRST (check_pending_read runs before this), so
+    # those branches get skipped and the tag rendered verbatim (the READ_LOCAL
+    # leak, then CRAWL_STARTED, then whatever the next new tag is).
+    #
+    # Strip-all-except-PLAY instead of an allowlist: every [[[TAG: ...]]] is a
```
omnigentx, line 206: **[nit — cosmetic] Strip-all-except-PLAY lookahead is narrow.**

`(?!PLAY[:\]])` only spares exactly `[[[PLAY:...]]]` / `[[[PLAY]]]` — a hypothetical `[[[PLAYBACK:...]]]` would be stripped. Harmless today (only `[[[PLAY:]]]` exists), just noting in case a new PLAY-prefixed tag is added later.

omnigentx:
## Review — story audio / crawl / voice / interrupt / passkey

**Verdict:** High quality, mostly **root-cause fixes** (not workarounds). 29 files, +2074/−772, 7 fixes bundled but each is its own commit. No blocker; one test gap and one process note.

### Verified good
- **CSRF passkey exempt** (`core/csrf.py`) — login flow, no token yet, WebAuthn challenge/signature is the defence, REGISTER stays protected. Well-reasoned.
- **Interrupt deadlock** (`pause_controller.py`) — `_turn_tasks` now tracks the whole turn (LLM + tool/delegate); `interrupt()`/`hard_kill()` no longer read the empty `_current_tasks` mid-delegate. Fixes the `_agent_lock` freeze that hung all later chats. Cleanup paths complete.
- **Audio false-end ~11s** (`tts.py`) — the old `is_generating(request_id)` never matched (pre-gen keys by `{story_title, chapter_file}`) so it always cancelled the requested chapter. Now the `.lock` on `cache_path` is the single source of truth; `X-TTS-Generating` pairs cleanly with the frontend HEAD-probe (underrun→resume vs real-end→advance).
- **Network retry cap** (`useAudioPlayer.js`) — kills the infinite 2s retry loop, resets on success.
- **STT keepalive/reconnect** (`_ws_streaming.py` + `soniox.py`) — 10s keepalive < 20s idle timeout; fatal (400/401/403→stop) vs transient (408→reconnect, no UI error flip). `wait_for(queue.get(), timeout)` is the right idle pattern.
- **Tag stripping / one playback path** (`chat.py`) — final sweep strips every `[[[TAG]]]` except `[[[PLAY:]]]`; `crawl_job_id` pulled out before stripping; `resolve_story_playback` shared by typed + voice.

### ⚠️ Needs attention
1. **Test gap on the 3 riskiest cross-layer changes.** Backend 51 + frontend 77 pass, but no dedicated test for: (a) interrupt **mid-delegate** deadlock (`_turn_tasks`) — highest priority; (b) passkey-authenticate CSRF exempt + REGISTER still 403 — cheapest; (c) STT fatal-vs-transient + keepalive reconnect. Crawl is manual-e2e on 3 sites only (not CI). Per the repo's integration/e2e rule these invariants should be pinned. Inline asks added on each.
2. **`_turn_tasks.setdefault` assumption** (`pause_controller.py:519`) — correct only if all `before_llm_call` in a turn share one asyncio task. Holds under fast-agent today, but a mid-delegate interrupt test would lock it down rather than trust it.
3. **Process — 7 fixes in one PR.** Commit-separated (good), but bundling CSRF + the ~1000-line crawl rewrite + audio + STT makes independent review/revert hard. Non-blocking; future PRs should split (esp. the crawl rewrite).

### Nit
- `chat.py` strip regex spares only exactly `[[[PLAY:]]]` — a future `[[[PLAY*]]]` tag would be stripped. Harmless today.

**Recommendation:** mergeable. Add tests for #1 (interrupt-deadlock + passkey CSRF first) before merge to avoid silent regressions. Happy to deep-review `crawl_poller.py` / `story_server.py` (self-heal, same-story guard, resume checkpoint) separately if you want.

omnigentx:
Addressed the two highest-priority test gaps (commit 1584682):

**1. Interrupt mid-delegate (highest priority)** — `test_interrupt_cancels_turn_even_mid_delegate` in `test_pause_controller.py`: drives the exact stuck path — `after_llm_call` clears `_current_tasks`, then asserts `interrupt()` still cancels the running turn task via `_turn_tasks` (and the asyncio task actually receives `CancelledError`). Plus `test_after_turn_complete_clears_turn_task_ref` so a stale ref can't be cancelled on a later turn. This pins the `setdefault`/one-task-per-turn assumption you flagged at `pause_controller.py:519`.

**2. Passkey CSRF boundary (cheapest)** — `TestPasskeyCsrfBoundary` in `test_csrf.py`: `authenticate/begin` + `finish` pass with a stale/missing CSRF cookie (the reported bug), while `register/begin` still 403s without a valid token and passes with a matching header. Both halves of the security boundary pinned.

73 passed across `test_pause_controller.py` + `test_csrf.py`.

Deferred (#3, STT fatal-vs-transient + keepalive): not yet covered — happy to add in a follow-up, or here if you'd prefer it before merge. Agreed on the process note too: the crawl rewrite should be its own PR next time.

omnigentx:
## Re-review of `1584682` — review comments addressed ✅

Verified by running the new tests in an isolated worktree (not just reading them).

| Review item | Status | Coverage |
|---|---|---|
| interrupt **mid-delegate** deadlock / `_turn_tasks` assumption (`pause_controller.py:519`) | ✅ Resolved | `test_interrupt_cancels_turn_even_mid_delegate` + `test_after_turn_complete_clears_turn_task_ref` — **passing** |
| passkey CSRF exempt + REGISTER still 403 (`csrf.py:89`) | ✅ Resolved | `TestPasskeyCsrfBoundary` (5 tests: authenticate exempt w/ stale & no cookie, register 403, register passes w/ matching header) — **passing** |
| STT fatal-vs-transient + keepalive reconnect (`soniox.py:111`) | ⏳ Deferred | consciously scoped out ("two highest-priority gaps") — lowest of the three |
| `tts.py` SSoT note, `chat.py` regex nit | — | informational only |

**Test quality:** real, not mock-of-the-subsystem — `test_interrupt_cancels_turn_even_mid_delegate` cancels a real asyncio task and asserts `CancelledError`; the CSRF tests use a real TestClient and pin both halves of the security boundary. Exactly the cross-layer coverage the review asked for.

**Run result:** the 7 review-related tests pass. A full run of both files showed `71 passed / 2 failed`, but the 2 failures (`test_find_pid_walks_uv_to_python_child_with_real_subprocess`, `test_uv_run_creates_two_level_process_tree`) are **unrelated to this PR** — they exercise `uv run` process-tree spawning, pass in a fully-provisioned checkout, and fail only in the half-set-up scratch worktree. This PR doesn't touch `_find_python_child`.

**Remaining:** only the STT fatal/transient unit test (#1c) is open. Reasonable to defer, but a small test feeding a fake 403 (assert `_fatal`) and a 408 (assert reconnect, no `error` emit) would close the last gap. Otherwise this LGTM.

PR #61, opened 2026-06-01
fix(team-template): address PR #58 ultrareview backlog (path-traversal + data-loss + nits)

### LINKED ISSUE #60: PR #60: address PR #58 ultrareview backlog (path-traversal + data-loss + 4 nits)
Backlog carried over from the **PR #58 ultrareview** (PR #58 merged at `5efa46e` before these were addressed — the merge commit only fixed the original 5 `/review` findings, so the 7 below now live on `main`). Tracking them as scope for **PR #60**.

Source: https://github.com/omnigentx/jarvis/pull/58 (inline comments + summary `issuecomment-4593415731`).

### 🔴 Should fix (normal — the dangerous class)

- [ ] **Path traversal via unvalidated `name:`** — `backend/services/team_template_rpc_handlers.py::_factory_yaml_path_for` builds the yaml path by raw string interpolation, bypassing `factory_svc._resolve()`. `name` comes from the yaml top-level `name:`, which `_validate_template_structure` never checks. An LLM (MCP `team_template_write_factory`, no human gate) can write `name: ../../target` → `reset_role`/`yaml_diff` `open()` files outside `team_templates/`. **Fix:** route through `factory_svc._resolve(name.replace('-','_'))` + catch `PathTraversalError`. Same pattern pre-exists in `routes/team_template.py::_resolve_yaml_for_session` — fix in lockstep.
- [ ] **sync+reload partial failure → silent overwrite** — `frontend/src/views/settings/SettingsRunningTemplates.vue`: if reset succeeds but reload throws, the catch never refetches; UI shows stale state, user re-edits and Saves over the just-synced DB. **Fix:** refetch in `finally` + tailored "sync applied but reload failed" message.
- [ ] **Switching role/session discards unsaved edits, no confirm** — `SettingsRunningTemplates.vue` watchers re-seed `draft` with no dirty-guard. **Fix:** mirror `SettingsYaml.vue::selectFile`'s `useConfirm()` guard on role-click, session-dropdown change, and Refresh.

### 🟡 Nit (batch in or defer)

- [ ] **`yaml_diff` AttributeError→500 on non-dict `team:`** — `team_template_rpc_handlers.py:223` missing `isinstance` guard (sibling `list_factory_templates` already guards it).
- [ ] **Reorder servers/skills stuck UNSAVED pill** — `dirty` is order-sensitive, `_buildPatch` is order-insensitive → pill stays lit, Save says "No changes". `SettingsRunningTemplates.vue:90`.
- [ ] **`.replace('-','_')` replaces only the first hyphen (JS) vs all (Python)** — modal shows wrong filename. `SettingsRunningTemplates.vue:707`. Display-only.
- [ ] **Dirty-dot `●` ignores `kind`** — cross-triggers on same-named files across the merged tree. `SettingsYaml.vue` (~L286). Display-only.

omnigentx:
## Summary

Resolves all **7 findings** from the PR #58 ultrareview (tracked in #60). PR #58 merged at `5efa46e` before these were addressed, so they landed on `main`.

### 🔴 Backend (the dangerous class)
- **bug_004 — path traversal.** `_factory_yaml_path_for` (rpc) and `_resolve_yaml_for_session` (route) built the yaml path by raw string interpolation, bypassing `_resolve()`. The template `name` is LLM-writable via `team_template_write_factory` and is **not** structurally validated, so `name: ../../x` could `open()` files outside `team_templates/`. Both now route through the new guarded `factory_svc.resolve_factory_path` → `PathTraversalError` → 400. Removed the orphaned `_YAML_DIR`.
- **bug_003 — `yaml_diff` 500.** A top-level list/scalar yaml or a non-dict `team:` made `.get()` raise `AttributeError` → unhandled 500. Both yaml_diff paths now guard the shape and 400 cleanly; a list `team:` falls back to top-level `roles:`.

### 🔴 Frontend (data-loss)
- **bug_011 — sync+reload partial failure.** Reset succeeded but reload threw → UI stayed stale → a re-edit+Save overwrote the just-synced DB. Reload failure is now a *partial success*: always refetch, distinct "synced to DB but reload failed" message.
- **bug_006 — discard unsaved edits.** Switching role/session or Refresh silently dropped in-progress edits. Added a `useConfirm` dirty-guard (mirrors `SettingsYaml.selectFile`) on role-click, session-select, and Refresh.

### 🟡 Nits
- **bug_008** — `dirty` was order-sensitive while `_buildPatch` sorts → reorder lit UNSAVED but Save said "No changes" (stuck pill). `dirty` now sort-normalizes servers/skills.
- **bug_001** — `.replace('-','_')` → `.replaceAll` so multi-hyphen template names show the correct factory filename.
- **bug_005** — dirty-dot `v-if` now checks `activeKind === f.kind` so a same-named team template doesn't light the ● on the fastagent row.

## Tests
- **Backend: 60 pass** (+4 new: path-traversal blocked on reset & yaml_diff; non-dict `team:` doesn't crash; top-level-list yaml → 400).
- **Frontend: `vite build` clean.** Added an e2e covering the bug_006 confirm-on-switch flow (cancel keeps the edit, confirm discards).

## Click-flow (bug_006 e2e)
Open Settings → Running Templates → edit a role's instruction → click another role → **"Discard unsaved changes?"** dialog → Cancel keeps the edit + UNSAVED pill → click again → Discard & Switch → switches, edit gone.

## Known gap
Frontend e2e could not run in the local sandbox (the app-boot harness fails for **unrelated** specs too, e.g. `chat-markdown-xss`); the new + existing specs are validated by CI.

Closes #60

### DIFF
--- a/backend/routes/team_template.py
+++ b/backend/routes/team_template.py
@@ -31,6 +31,7 @@ from sqlalchemy.orm import Session
 
 from core.auth import verify_api_key
 from core.database import get_db
+from services import team_template_factory_service as factory_svc
 from services import team_template_service as svc
 from services.team_reload import reload_roles
 
@@ -38,23 +39,17 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter(prefix="/api/team-sessions", tags=["team-template"])
 
-# Yaml templates dir — used by /reset/{role} to look up the factory default.
-_YAML_DIR = Path(__file__).resolve().parent.parent / "team_templates"
-
 
 def _resolve_yaml_for_session(session_id: str) -> Path:
     """Locate the yaml template for a team session.
 
     Reads ``team_sessions.template.name`` to find which yaml the team was
-    created from (e.g. 'agile-team' → 'agile_team.yaml').
+    created from (e.g. 'agile-team' → 'agile_team.yaml'). Delegates to the
+    guarded ``factory_svc.resolve_factory_path`` so a malicious template name
+    can't escape ``team_templates/`` (raises ``PathTraversalError``).
     """
     template = svc.get_template(session_id)
-    name = template.get("name") or "agile-team"
-    candidate = _YAML_DIR / f"{name.replace('-', '_')}.yaml"
-    if not candidate.exists():
-        # Fallback: try literal name
-        candidate = _YAML_DIR / f"{name}.yaml"
-    return candidate
+    return factory_svc.resolve_factory_path(template.get("name") or "agile-team")
 
 
 class PatchRoleBody(BaseModel):
@@ -241,7 +236,10 @@ async def reset_role(
     Writes audit rows for every field that diverged. Other roles untouched.
     """
     edited_by = _principal(request)
-    yaml_path = _resolve_yaml_for_session(session_id)
+    try:
+        yaml_path = _resolve_yaml_for_session(session_id)
+    except factory_svc.PathTraversalError as e:
+        raise HTTPException(status_code=400, detail=str(e))
     try:
         result = svc.reset_role_to_yaml(
             db,
@@ -284,7 +282,10 @@ async def yaml_diff(session_id: str):
     except svc.NotFoundError as e:
         raise HTTPException(status_code=404, detail=str(e))
 
-    yaml_path = _resolve_yaml_for_session(session_id)
+    try:
+        yaml_path = _resolve_yaml_for_session(session_id)
+    except factory_svc.PathTraversalError as e:
+        raise HTTPException(status_code=400, detail=str(e))
     if not yaml_path.exists():
         raise HTTPException(
             status_code=404,
@@ -293,7 +294,16 @@ async def yaml_diff(session_id: str):
 
     with yaml_path.open(encoding="utf-8") as f:
         yaml_doc = yaml_lib.safe_load(f) or {}
-    yaml_template = yaml_doc.get("team") or yaml_doc
+    # Guard the shape before calling .get — a top-level list/scalar yaml, or a
+    # truthy non-dict ``team:`` (``team: [pm, dev]``), would otherwise raise
+    # AttributeError → unhandled 500.
+    if not isinstance(yaml_doc, dict):
+        raise HTTPException(
+            status_code=400,
+            detail=f"factory yaml at {yaml_path} is not a mapping",
+        )
+    team = yaml_doc.get("team")
+    yaml_template = team if isinstance(team, dict) else yaml_doc
     yaml_roles = yaml_template.get("roles") or {}
     current_roles = current_template.get("roles") or {}
 
--- a/backend/routes/team_template.py
+++ b/backend/routes/team_template.py
@@ -275,8 +275,6 @@ async def yaml_diff(session_id: str):
     Decision 2026-05-17: yaml is FACTORY DEFAULT, not continuous source.
     This endpoint lets the user *see* drift without enforcing convergence.
     """
-    import yaml as yaml_lib
-
     try:
         current_template = svc.get_template(session_id)
     except svc.NotFoundError as e:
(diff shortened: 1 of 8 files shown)


omnigentx:
## Code review — verified locally (checked out branch, ran tests: 19 passed in `test_team_template_rpc_handlers.py`, incl. the 4 new ones)

Solid PR — every fix carries a *why* comment, the two dangerous classes (path-traversal + reset/reload data-loss) are root-caused. **Approve with 3 non-blocking suggestions.**

### Correctness
- ✅ **bug_004 path-traversal** — confirmed the guard at the leaf `_resolve()` (`team_template_factory_service.py:80-85`): rejects empty / `/` / `\` / leading-`.` names, then re-checks `_FACTORY_DIR in candidate.parents`. `resolve_factory_path` tries both underscore and literal forms but **both go through `_resolve`** — no bypass. `../../secrets` and absolute paths are blocked.
- ⚠️ **bug_003 — subtle semantics change for empty `team: {}`.** Old `yaml_doc.get("team") or yaml_doc` fell back to top-level on any *falsy* team. New `team if isinstance(team, dict) else yaml_doc`:
  - `team: [pm, dev]` (list) → falls back to top-level `roles:` ✓ (the intended fix)
  - `team: {}` (empty dict, falsy) → old code fell back to top-level `roles:`, **new code keeps `{}` → empty roles.**

  Very narrow (a `team: {}` + root `roles:` doc is unusual), but it's inconsistent with bug_003's own stated intent ("unusable team → fall back"). If that's the intent, prefer `team if (isinstance(team, dict) and team) else yaml_doc`; if the dict/non-dict distinction is deliberate, a one-line comment would stop a future reader from "fixing" it.

### Maintainability (SSoT)
- ⚠️ **Duplicate guard logic across two files.** Both the shape-guard (`isinstance(yaml_doc, dict)` → 400, `team if isinstance(team, dict)`) and the `PathTraversalError` → 400 translation are now copy-pasted in `routes/team_template.py` **and** `services/team_template_rpc_handlers.py`. Classic two-sources-that-will-drift. Suggest hoisting into `factory_svc`, e.g. a `load_factory_roles(path)` that raises `ValidationError` on a non-mapping — callers just map exception→status.

### Test coverage
- ⚠️ **Route layer has the new guards but no new test.** All 4 new tests are at the `rpc_handlers` layer. `routes/team_template.py` got identical `reset_role`/`yaml_diff` → 400 guards, but no diff test proves the HTTP layer translates `PathTraversalError`→400. Because the logic is duplicated (above), only half is covered. `tests/test_routes/test_team_template_routes.py` already exists — one TestClient traversal test there closes the other half cheaply.
- Frontend e2e known-gap (sandbox boot fails for unrelated specs too) is declared honestly and leans on CI — fine.

### Frontend — all correct
- ✅ **bug_008** dirty sort-normalize matches `_buildPatch`; spread-before-sort avoids mutating `live`.
- ✅ **bug_006** switching `<select>` from `v-model` → `:value`/`@change` is the *right* move to insert an async confirm; revert-on-cancel works because two-way bind is gone.
- ✅ **bug_011** partial-success is a proper root-cause fix for the data-loss. Minor asymmetry: success path `await loadTemplate`, outer-catch uses fire-and-forget — acceptable on the error path.

### Security
No new hole. Pre-existing, out-of-scope note: a name with an embedded null byte passes `_resolve` (no `/`, `\`, no leading `.`) and `open()` then raises `ValueError` → 500 instead of 400. Not introduced here; flag only if you later tighten.

---
**Suggestions (none blocking):** (1) clarify/align the empty-`team: {}` fallback, (2) hoist the shape-guard+resolve into one `factory_svc` helper to kill the route↔rpc duplication, (3) add a route-layer traversal→400 test.

omnigentx:
Thanks for the review — all 3 non-blocking suggestions addressed in `b401316`:

1. **empty `team: {}` fallback** — fixed. The guard is now `team if (isinstance(team, dict) and team) else <top-level>`, so an empty/non-dict `team:` falls back to top-level `roles:` (matching bug_003's "unusable team → fall back" intent). Regression test added.
2. **SSoT — hoisted the shape-guard.** New `factory_svc.load_factory_roles(path)` does the open + unwrap-roles + non-mapping→`ValidationError` once; both `yaml_diff` callers (route + rpc) now delegate. The `PathTraversalError`/`ValidationError` → status mapping stays per-caller on purpose — the route raises `HTTPException`, the rpc handler returns a `{error,status}` envelope, so they can't share that line.
3. **route-layer test** — added traversal→400 tests for both `yaml_diff` and `reset` using the real (un-mocked) `_resolve_yaml_for_session`.

Also noted the embedded-null-byte → 500 edge as out-of-scope/pre-existing; left untouched.

+7 tests (5 `load_factory_roles` incl. the empty-team regression, 2 route traversal). 67 team-template tests pass; CI green.

PR #73, opened 2026-06-04
feat(voice): realtime barge-in + WebRTC audio (iOS echo fix)

omnigentx:
## What & why

Fixes two voice problems and lays the proper transport for both:

1. **Barge-in latency (desktop):** TTS kept playing when you talked over it, only stopping once your transcript submitted. Root cause: the server flipped `bot_speaking=False` the instant *synthesis* finished, but the client buffers seconds of audio ahead — so a barge-in during the playback tail was a silent no-op.
2. **iPhone Safari echo/feedback loop:** the bot heard its own TTS and transcribed it as a new turn. Root cause: iOS Safari runs acoustic echo cancellation (AEC) only for audio played through a WebRTC `<audio>` track, never for Web Audio API playback — which the old WS+AudioWorklet path used.

## Approach (phased)

- **Phase 0 — barge-in SSoT:** `bot_speaking` now tracks *real client playback* (until `playback_done`), not synthesis. `_cancel_inflight` flushes even with no active task. Client reports `playback_done` via a unit-tested `playbackDoneTracker`. → instant barge-in on AEC-capable browsers (Chrome).
- **Phase 1 — WebRTC audio transport:** audio (mic + TTS) now runs over an `RTCPeerConnection` so the browser AEC scrubs the bot out of the mic (the iOS fix). Control/STT/barge-in stay on `/ws/voice`. Backend uses `aiortc` (`services/webrtc_voice.py`); TTS plays via a hidden `<audio playsinline>`. STUN for NAT (`JARVIS_WEBRTC_ICE` to add TURN).
- **Fail loud, no silent fallback:** WebRTC is required; failures surface an error. The legacy WS audio path is only reachable by an explicit opt-in (`localStorage.voice_transport='ws'`) for A/B debugging.

`feed_audio` (and therefore **Soniox / faster-whisper / gipformer** STT) is unchanged — it still receives 16 kHz mono s16 regardless of transport. The browser↔server hop is the only thing that moved.

## Tests (all green, headless)

- backend: `test_services` + `test_ws_voice` **806** + voice **18** (aiortc loopback both directions; `webrtc_offer→answer` over the real WS handler; onset / playback-tail / `playback_done` barge-in).
- frontend: unit **153** (incl. `playbackDoneTracker` 7); e2e **4** — `voice-webrtc.spec.ts` (real `webrtc_offer` from production code), `voice-barge-in.spec.ts` (WS-path SSoT ×2), dictation regression.

## ⚠️ Needs manual / on-device validation

Audio fidelity and the actual iOS AEC **cannot** be checked headless. See **`docs/VOICE_WEBRTC_TESTING.md`**:
1. **Mac Chrome / localhost first** (no deploy) — validates ~95%: WebRTC connects, `/ws/voice` carries JSON only, barge-in stops TTS instantly.
2. **iPhone Safari (prod / HTTPS)** — confirms no echo loop. If it can't connect over NAT, open UDP or add a TURN server.

## Commits
- barge-in SSoT (BE) · playback_done tracker + e2e (FE) · WebRTC bridge + loopback · wire WebRTC into the session (BE+FE) · fail-loud (no silent fallback)

### DIFF
--- a/backend/routes/ws_voice.py
+++ b/backend/routes/ws_voice.py
@@ -61,6 +61,7 @@ from services.sse_progress import (
     merge_hooks,
     progress_manager,
 )
+from services.webrtc_voice import WebRtcVoiceSession
 
 
 def _chunk_rms_peak(pcm_chunk: bytes) -> tuple[int, int]:
@@ -302,6 +303,13 @@ async def voice_ws(ws: WebSocket) -> None:
     tts_task: Optional[asyncio.Task] = None
     user_query_pending = False  # set when STT finalises text but agent task hasn't started yet
     bot_speaking = False        # True from tts_start until client playback_done / barge-in
+    # WebRTC audio transport. When the client negotiates it (``webrtc_offer``),
+    # mic audio arrives on the RTCPeerConnection track (→ feed_audio) and TTS
+    # goes out the same connection — so the browser's AEC scrubs the bot's voice
+    # out of the mic (the iOS Safari echo fix). Signalling/control/STT/barge-in
+    # stay on this WS. Falls back to the WS audio path when not negotiated.
+    webrtc_session = None
+    webrtc_active = False
     # Dictation mode: client picked the bottom mic in ChatInput (press-to-talk
     # transcription) instead of the top hands-free button. In this mode the
     # STT pipeline runs as usual but ``final_transcript`` does NOT trigger
@@ -347,6 +355,11 @@ async def voice_ws(ws: WebSocket) -> None:
         # until I stop talking and the transcript submits".
         if cancelled or bot_speaking:
             bot_speaking = False
+            # WebRTC: also drop any TTS audio still queued on the outbound track
+            # so the bot goes silent immediately (cancelling the task stops
+            # *producing* audio, but already-queued frames would keep playing).
+            if webrtc_session is not None:
+                webrtc_session.tts_track.flush()
             logger.info(
                 "[ws_voice] barge-in (%s) cancelled %s",
                 reason, ",".join(cancelled) or "playback-only",
@@ -540,25 +553,44 @@ async def voice_ws(ws: WebSocket) -> None:
         await out_queue.put({"type": "tts_start"})
         bytes_sent = 0
         chunks_sent = 0
+        # Route audio over WebRTC when the client negotiated it (the browser
+        # plays the TTS track via <audio>, which is what lets its AEC scrub the
+        # echo on iOS). Otherwise the WS binary path. tts_start/tts_end JSON
+        # still go over the WS either way so the UI updates the same.
+        use_webrtc = webrtc_active and webrtc_session is not None
+
+        async def _emit(pcm: bytes) -> None:
+            nonlocal bytes_sent, chunks_sent
+            bytes_sent += len(pcm)
+            chunks_sent += 1
+            if use_webrtc:
+                webrtc_session.send_pcm(pcm)
+            else:
+                await out_queue.put(pcm)
+
         logger.info(
-            "[ws_voice] speak() start: provider=%s path=%s text_len=%d",
-            type(provider).__name__, path, len(text),
+            "[ws_voice] speak() start: provider=%s path=%s text_len=%d transport=%s",
+            type(provider).__name__, path, len(text), "webrtc" if use_webrtc else "ws",
         )
         try:
             if stream_pcm is not None:
                 async for pcm in stream_pcm(text):
-                    bytes_sent += len(pcm)
-                    chunks_sent += 1
-                    await out_queue.put(pcm)
+                    await _emit(pcm)
             else:
                 async for pcm in _mp3_stream_to_pcm(provider.stream_audio(text)):
-                    bytes_sent += len(pcm)
-                    chunks_sent += 1
-                    await out_queue.put(pcm)
+                    await _emit(pcm)
             logger.info(
                 "[ws_voice] speak() done: chunks=%d bytes=%d",
                 chunks_sent, bytes_sent,
             )
+            if use_webrtc:
+                # WebRTC: the server paces the track, so we know exactly when the
+                # user stops hearing the bot — wait for the track to drain, then
+                # drop bot_speaking. This is the WebRTC analogue of the WS path's
+                # client ``playback_done``. tts_task stays alive for the whole
+                # playback, so an onset barge-in cancels it mid-tail for free.
+                await webrtc_session.tts_track.wait_drained()
+                bot_speaking = False
         except asyncio.CancelledError:
             await out_queue.put({"type": "barge_in_ack"})
             raise
@@ -848,6 +880,12 @@ async def voice_ws(ws: WebSocket) -> None:
             if msg["type"] == "websocket.disconnect":
                 return
             if "bytes" in msg and msg["bytes"] is not None:
+                # When WebRTC is negotiated the mic arrives on the RTCPeerConnection
+                # track (pumped into feed_audio there), so ignore any WS PCM frames
+                # to avoid feeding STT twice. The client stops sending them, but
+                # guard server-side too.
+                if webrtc_active:
+                    continue
                 # Diagnostic: classify each frame as "during bot" vs "silent
                 # baseline" so we can compare RMS distributions and confirm
                 # whether browser AEC is actually scrubbing the bot's voice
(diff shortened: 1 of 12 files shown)


frontend/src/composables/useVoiceSession.js
```diff
@@ -305,8 +330,33 @@ function _createVoiceSession() {
...
+        case 'tts_start':
+          status.value = 'speaking'
+          wasInterrupted.value = false
+          playbackTracker.ttsStart()   // re-arm playback_done for this turn
+          break
         case 'tts_end':
+          // Server finished synthesising. Playback usually continues a few more
+          // seconds; the tracker fires playback_done once the queue drains (or
+          // right now if a very short reply already drained before tts_end).
+          playbackTracker.ttsEnd(playbackSources.size)
```
omnigentx: **HIGH — likely bug, SSoT violation: onset barge-in dies during the WebRTC playback tail.**

This `playbackTracker.ttsEnd(...)` is **not gated on `webrtcMode`**. In WebRTC mode `_enqueueAudio` is bypassed (the `!webrtcMode && Blob` guard further up), so `playbackSources` is **always empty** → `ttsEnd(0)` makes the tracker fire `playback_done` **immediately** at `tts_end`.

Server then sets `bot_speaking = False` (`backend/routes/ws_voice.py:962`) *while the WebRTC track is still playing* — `speak()` is still in `await webrtc_session.tts_track.wait_drained()` (`ws_voice.py:589-593`).

The onset barge-in gate is purely `if bot_speaking:` (`ws_voice.py:395`). So once the premature `playback_done` lands, a user talking over the still-playing buffered audio is a **no-op** — the bot keeps playing. That is the exact bug this PR set out to fix, reintroduced on the WebRTC path (the default/primary path).

Root cause: two writers to `bot_speaking` with different timing. In WebRTC mode the **server `wait_drained()` is the authority**; the client must not send `playback_done`.

Suggested fix — gate the tracker in WebRTC mode (and likewise the `chunkEnded`/`ttsStart` wiring):
```js
case 'tts_end':
  if (!webrtcMode) playbackTracker.ttsEnd(playbackSources.size) // WebRTC: server owns the drain edge
  status.value = ...
```

**Missing coverage:** no test exercises onset barge-in during a WebRTC playback tail — the loopback test only checks offer/answer + resample plumbing, not the `bot_speaking` SSoT on this path. Please add a backend test that catches this race.

backend/routes/ws_voice.py
```diff
@@ -855,8 +915,51 @@ async def _handle_user_turn(text: str) -> None:
...
                     _cancel_inflight("client_barge_in")
+                elif kind == "playback_done":
+                    # Authoritative "the user has stopped hearing the bot"
+                    # signal: the client finished playing every buffered TTS
+                    # chunk. Until this lands, bot_speaking stays True (set at
+                    # tts_start, kept past synthesis end) so a barge-in during
+                    # the playback tail still flushes the client queue. This is
+                    # the SSoT fix: server ``bot_speaking`` mirrors real client
+                    # playback instead of server-side synthesis state.
+                    bot_speaking = False
```
omnigentx: Backend side of Finding 1: this `bot_speaking = False` is correct for the **WS** path (client `playback_done` = real drain). But in **WebRTC** mode the client's `playback_done` arrives at `tts_end` (client has no local buffer), so this fires too early and races the authoritative `wait_drained()` at line 593. Either ignore `playback_done` when `webrtc_active` here, or (preferred) stop the client from sending it in WebRTC mode. See the detailed note on `useVoiceSession.js:359`.

omnigentx (commented):
Code review — 4 findings (1 HIGH likely-bug, 1 MEDIUM, 2 LOW). Posted inline. Highlights:

- **HIGH / SSoT (Finding 1):** in WebRTC mode the client still fires `playback_done` at `tts_end` (its local queue is always empty there), prematurely lowering server `bot_speaking` while the WebRTC track is still playing → **onset barge-in becomes a no-op during the WebRTC playback tail** — the exact bug this PR fixes, reintroduced on the primary path. Two writers to `bot_speaking` with conflicting timing. Needs a gate + a test.
- **MEDIUM (Finding 2):** client `webrtc_error` handler leaves the half-open `RTCPeerConnection` + hidden `<audio>` un-torn-down.
- **LOW (Finding 3):** frontend STUN is hardcoded while backend is configurable (`JARVIS_WEBRTC_ICE`) — no way to point FE at a TURN server (the symmetric-NAT / iPhone-on-cellular case).
- **LOW/nit (Finding 4):** `_iceGatheringComplete` never `clearTimeout`s.

Verified safe: `av` 17.0.1 → 16.1.0 downgrade is forced by aiortc and no application code imports `av` directly. Nice work on the why-comments and the offer/answer-over-real-WS test.

frontend/src/composables/useVoiceSession.js
```diff
@@ -305,8 +330,33 @@ function _createVoiceSession() {
...
+          // Server accepted the WebRTC offer — complete the handshake; the
+          // connection forms once ICE finishes and audio starts flowing on the
+          // media track (mic out, TTS in).
+          if (pc.value && msg.sdp) {
+            pc.value
+              .setRemoteDescription({ type: msg.sdp_type || 'answer', sdp: msg.sdp })
+              .catch((e) => console.warn('[voice] setRemoteDescription failed', e))
+          }
+          break
+        case 'webrtc_error':
```
omnigentx, line 343: **MEDIUM:** on `webrtc_error` we set `status='error'` but never tear down the peer connection. By this point `_startWebRtc` has already set `webrtcMode = true` and left `pc.value` open plus the hidden `<audio>` attached to `document.body`. It only gets cleaned up on the next mic toggle (`stop()` → `_teardownWebRtc()`), so a stale PC + DOM node linger. Call `_teardownWebRtc()` here (and on `connectionState === 'failed'`).

frontend/src/composables/useVoiceSession.js
```diff
@@ -402,13 +453,93 @@ function _createVoiceSession() {
...
+
+  /**
+   * Negotiate WebRTC audio over the already-open WS. Mic goes out as a track;
+   * the bot's TTS comes back as a track we play through a hidden <audio
+   * playsinline> — which is what engages the browser AEC (incl. iOS Safari).
+   * On any failure we leave webrtcMode false so start() uses the WS audio path.
+   */
+  async function _startWebRtc(sock, ms) {
+    const peer = new RTCPeerConnection({
+      iceServers: [{ urls: 'stun:stun.l.google.com:19302' }],
```
omnigentx: **LOW:** STUN is hardcoded here, but the backend ICE list is configurable via `JARVIS_WEBRTC_ICE` (`services/webrtc_voice.py:191`). The frontend has no way to point at a TURN server — which is exactly what the symmetric-NAT / iPhone-on-cellular case (the scenario this PR targets) needs. The testing doc acknowledges this ("tell me and I'll wire it through"). Consider sourcing FE `iceServers` from a config endpoint before on-device prod testing, so a TURN deployment doesn't require a code change.

frontend/src/composables/useVoiceSession.js
```diff
@@ -402,13 +453,93 @@ function _createVoiceSession() {
...
+  // Wait for ICE gathering to finish (non-trickle) so the offer SDP carries all
+  // candidates — simplest signalling over our WS. Cap the wait so a browser that
+  // never reports 'complete' still sends the offer with whatever it gathered.
+  function _iceGatheringComplete(peer, timeoutMs = 2500) {
+    if (peer.iceGatheringState === 'complete') return Promise.resolve()
+    return new Promise((resolve) => {
+      const done = () => { peer.removeEventListener('icegatheringstatechange', check); resolve() }
+      const check = () => { if (peer.iceGatheringState === 'complete') done() }
+      peer.addEventListener('icegatheringstatechange', check)
+      setTimeout(done, timeoutMs)
```
omnigentx: **LOW / nit:** if ICE gathering completes before `timeoutMs`, this `setTimeout` still fires `done()` a second time. Harmless (promise already resolved, `removeEventListener` is idempotent), but cleaner to `clearTimeout` once `done()` runs.

omnigentx:
## Review addressed — pushed `4d52f6e`

Thanks, the HIGH finding was a real bug I introduced on the default path. Fixed all five:

**🔴 HIGH — onset barge-in dies in WebRTC playback tail (SSoT).** Correct diagnosis: `playbackSources` is always empty in WebRTC mode, so `ttsEnd(0)` fired `playback_done` at `tts_end` → server cleared `bot_speaking` mid-playback.
- FE: gated the tracker on `!webrtcMode` (`ttsStart`/`ttsEnd`). In WebRTC the server's `wait_drained()` is the sole drain authority.
- BE (defensive): `playback_done` is ignored when `webrtc_active`.
- **Test added:** `voice-webrtc.spec.ts` now drives a full `tts_start`→`tts_end` turn and asserts the client sends **no** `playback_done` in WebRTC mode.

**🟠 MEDIUM — stale PC/`<audio>` on error.** `webrtc_error` and `connectionState === 'failed'` now call `_teardownWebRtc()` immediately.

**🟡 LOW — FE STUN hardcoded, no TURN.** Added `GET /api/voice/ice` (serves `iceServers` parsed from `JARVIS_WEBRTC_ICE`, incl. `turn:user:pass@host`); the frontend fetches it before negotiating, falling back to public STUN. Adding a TURN relay for prod/iPhone is now an env change. Unit-tested `parse_ice_servers` (TURN creds, `:` in password, default).

**🟡 LOW — `clearTimeout` nit.** Done.

Verified: backend webrtc+ws_voice **19/19**, frontend voice e2e **4/4**, build green.

One honest note on the SSoT test request: a *backend* test that exercises the WebRTC `bot_speaking` race at the WS level isn't feasible deterministically — `speak()`'s `wait_drained()` only advances when a real peer consumes the track, which the TestClient can't provide (it'd hang). The frontend e2e (client never sends `playback_done`) + the backend guard cover both writers instead.

PR #78, opened 2026-06-06
fix(tts): cap EdgeTTS chunks at 500 + per-chunk retry

omnigentx:
## Symptom
Story chapter audio on app.omnigentx.com played intermittently — *tậm tịt, lúc được lúc không*. Chat TTS stayed fine even while pre-gen ran.

## Root cause (from prod logs)
Both TTS paths sent a **single oversized `edge_tts` request** per chapter:
- `EdgeTTSProvider._split_tiered` split text into tiers `(50, 300)` then dumped **all the rest into one giant chunk** → a ~9k-char chapter became one ~8.6k-char request.
- `tts_pregen_job` sent the whole chapter text in one `edge_tts.Communicate`.

A single oversized request streams **slower than real-time playback** (prod: ~3.5 MB over 5–6 min, 3 "chunks") and intermittently returns `No audio received` / truncates → the file is removed → silent chapter. Short chat replies fit in tier-1/2 (small, reliable requests), so chat never hit this — confirming it was **not** an IP throttle or pre-gen overload. (Verified: server egress to MS CDN was 14 MB/s; only the EdgeTTS WSS trickled.)

## Fix
- **`EDGE_MAX_CHUNK = 500`** — `_split_tiered` keeps tier-1/2 tiny for low TTFB, then caps every remaining chunk at 500 chars (sentence/comma/whitespace boundaries). No more giant remainder.
- **`_synth_chunk`** — each chunk synthesized with **retry + linear backoff**, buffered so a retried attempt can't duplicate audio; fails loud if a chunk yields no audio after retries (caller drops the partial file → no silent gap).
- **`tts_pregen_job`** — reuses `_split_tiered` + the same per-chunk retry, preserving pause/cancel (`CancelledError` re-raised, never retried).

## Tests
- New `test_tts_chunking.py`: every chunk ≤ 500, no text dropped, retry-then-succeed, empty-after-all-attempts, cancel-not-retried, `stream_audio` fails loud on unrecoverable chunk.
- Updated the old "exactly 3 tiers" assertion → "bounded chunks".
- **813 passed** in `test_services/`.

### DIFF
--- a/backend/services/tts.py
+++ b/backend/services/tts.py
@@ -29,6 +29,12 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_EDGE_VOICE = "vi-VN-NamMinhNeural"
 DEFAULT_EDGE_RATE = "+20%"
+# Largest text size sent in ONE edge_tts request. A single oversized request
+# (e.g. a whole ~9k-char story chapter) streams slower than real-time playback
+# and intermittently returns "no audio" / truncates — which made long story
+# chapters play "tậm tịt". Capping every chunk keeps each request small, fast
+# and reliable. Tier 1/2 stay far below this for low time-to-first-byte.
+EDGE_MAX_CHUNK = 500
 
 
 class TTSProvider(ABC):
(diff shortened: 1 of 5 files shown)


omnigentx:
## Review — blocking bug + missing integration coverage

The chunk-cap (≤500), per-chunk retry, and 30s per-chunk timeout are correct and well-reasoned. But there is one **blocking correctness bug** plus a **test-coverage gap** that let it through.

### 🔴 Blocking — pre-gen leaves a truncated chapter that is then served forever

`backend/services/tts_pregen_job.py:293`
```python
except Exception as e:
    ...
    if os.path.exists(cache_path) and bytes_written == 0:   # only when ZERO bytes
        os.remove(cache_path)
```

Failure sequence with the new code:
1. Chunks 1–2 succeed → `bytes_written > 0`, partial file on disk.
2. A later chunk fails after retries / times out (3 × 30s) → `raise RuntimeError` (`:246`).
3. The `except` runs but **skips removal** because `bytes_written != 0`.
4. The lock file *is* removed (`:295`).
5. Next run hits the cache-hit short-circuit at `:176` (`os.path.exists(cache_path) and not lock`) → returns `True`, treating the **truncated file as a complete chapter**. It is never regenerated.

Net: a mid-chapter chunk failure produces a permanently cached, silently truncated chapter — exactly the "tậm tịt" symptom this PR exists to fix. The new 30s timeout actually *widens* this path (a stalled chunk now reliably raises mid-chapter). The streaming path is fine — `backend/routes/tts.py:238` removes the partial on any exception. Only pre-gen is wrong; the `bytes_written == 0` guard was written for the old streaming model.

**Fix (preferred, root-cause):** write to `cache_path + ".tmp"` and `os.replace()` to the final path only after all chunks succeed. A present `cache_path` then always means "complete", even across crash / SIGKILL. **Minimal:** drop the `and bytes_written == 0` condition.

### 🟡 No integration/e2e coverage on the path where the bug lives

- `test_tts_chunking.py` fully fakes `edge_tts.Communicate`, so it only unit-tests `_split_tiered` / `_synth_chunk`. Good, but it never exercises `execute_task`'s real file I/O, `bytes_written` accounting, cleanup branch, or the cache-hit short-circuit.
- `test_tts_split.py::test_pregen_job_still_uses_edge_directly` is a static source-string grep, not a runtime test.
- `tests/e2e/` has no story-audio / pre-gen flow.

Two integration tests needed (fake `edge_tts` only at the network boundary, let file I/O run for real):
1. **Happy path:** long chapter → `execute_task()` → assert `cache_path` exists and content == all chunks concatenated (nothing dropped/duplicated).
2. **Failure path:** chunks 1–2 ok, chunk 3 empty after retries → assert `cache_path` does **not** exist. This test fails against current code — that is the proof the bug is live.

### Verdict
Timeout + chunk-cap work is solid and worth keeping. Please (1) fix pre-gen cleanup (temp + atomic rename) and (2) add the two integration tests before merge.

omnigentx:
Đã fix cả 2 điểm ở commit `3283e27`.

### 🔴 Atomic publish — không bao giờ để file chương cụt trong cache
Đúng bug. Pre-gen giờ ghi vào `cache_path + ".tmp"` và `os.replace()` sang `cache_path` **chỉ sau khi tất cả chunk thành công**. Một `cache_path` tồn tại ⇒ **luôn** là chương hoàn chỉnh (kể cả khi crash/SIGKILL giữa chừng). Cả 2 nhánh cleanup (`CancelledError`, `Exception`) xóa file `.tmp`; đã bỏ guard `bytes_written == 0` (vốn viết cho mô hình stream cũ).

### 🟡 Integration test trên đúng path có bug
`test_tts_pregen_integration.py` — chạy `execute_task()` với **file I/O thật**, chỉ fake `edge_tts` ở biên mạng:
1. **Happy path** → `cache_path` tồn tại, nội dung == các chunk nối lại (không mất/lặp), `.tmp`/`.lock` đã dọn.
2. **Failure path** (chunk 1–2 ok, chunk 3 rỗng sau retry) → `assert not os.path.exists(cache_path)`. Test này **fail với code cũ** (file cụt sót lại) → đúng bằng chứng bug đã được sửa.

**Green:** 816 passed trong `test_services/`.

Cảm ơn review — bug này tôi bỏ sót thật, đặc biệt timeout 30s làm path raise-giữa-chương dễ xảy ra hơn.

PR #82, opened 2026-06-10
feat(voice): Cloudflare TURN credentials for WebRTC NAT traversal

omnigentx:
## Why

Voice over WebRTC has **no ICE path from any network outside the tailnet**. The deployment has zero public UDP inbound (web traffic rides a cloudflared tunnel; the host sits behind home NAT + docker bridge NAT). Evidence from prod logs (iPhone Safari on 5G, 2026-06-10):

```
aioice: Check CandidatePair(('172.19.0.3', 54772) -> ('215.67.92.112', 52149)) -> FAILED
aioice: Check CandidatePair(('172.19.0.3', 54772) -> ('171.244.120.115', 7806)) -> FAILED
aioice: ICE failed
```

Desktop Chrome only ever worked because the Mac runs Tailscale, whose interface contributes a candidate the server can route to (`100.114.26.107` — confirmed as the only SUCCEEDED pair in prod logs). Industry-standard fix for this topology: a TURN relay.

## What

- `get_ice_servers()` in `services/webrtc_voice.py` — the **single source of truth** for ICE config, consumed by both `/api/voice/ice` (browser) and the server's own `RTCPeerConnection`:
  1. Cloudflare Realtime TURN creds, minted on demand from `JARVIS_CF_TURN_KEY_ID` + `JARVIS_CF_TURN_API_TOKEN` (`POST /v1/turn/keys/{id}/credentials/generate-ice-servers`, TTL 24 h)
  2. fallback: existing `JARVIS_WEBRTC_ICE` static list
  3. fallback: public STUN (unchanged default)
- Mint cache refreshes at half-life (12 h) so creds handed to a new session always outlive it; a stale-but-valid batch is served if the CF API blips (loud `logger.exception`, no silent fallback). Hard expiry at TTL−10 min.
- `WebRtcVoiceSession(ice_servers=…)` now accepts browser-shaped dicts (urls list + username/credential) as well as plain url strings; the ws handler pre-fetches via `asyncio.to_thread` so a cold mint (blocking HTTPS ≤10 s) never stalls the event loop.
- `backend/.env.example` documents the new vars.

With the CF vars **unset**, behavior is byte-identical to before (STUN-only) — safe to deploy ahead of credential provisioning.

## Click-flow

iPhone Safari (5G/CGNAT) → app.omnigentx.com/chat → mic on → frontend fetches `/api/voice/ice` (now contains `turn.cloudflare.com` + minted creds) → `webrtc_offer` over WS → both peers allocate relay candidates on Cloudflare's edge → ICE succeeds via relay → mic RTP reaches `pump_inbound_track` → Soniox STT → transcripts.

## Tests

- `tests/test_services/test_webrtc_voice.py` — **7 passed**: CF mint happy path (URL/auth-header/TTL asserted), cache hit, half-life re-mint, stale-serve on API failure + hard-expiry fallback, dict/str `ice_servers` shapes, and the existing aiortc loopback e2e (real RTCPeerConnection pair, no mocks of the changed subsystem).
- `tests/test_routes/test_ws_voice.py` — **17 passed** (webrtc_offer signalling path included).
- Real-relay e2e (forced `relay` ICE policy through live Cloudflare TURN) requires the TURN key — will run post-provisioning with the aiortc client harness used during diagnosis.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -72,3 +72,15 @@ CORS_ORIGINS=*
 # Per-user fork URL (e.g. https://github.com/<your-username>/jarvis) is normally
 # stored in the encrypted DB via Settings → Services → jarvis_repo, NOT here.
 # JARVIS_REPO_URL=
+
+# ─── WebRTC TURN (voice from CGNAT/cellular networks) ────────────────────────
+# This deployment has no public UDP inbound (web rides a cloudflared tunnel),
+# so WebRTC voice from outside the LAN/tailnet (e.g. iPhone on 5G) needs a
+# TURN relay. Recommended: Cloudflare Realtime TURN — create a TURN app at
+# dash.cloudflare.com → Realtime → TURN, then set BOTH:
+# JARVIS_CF_TURN_KEY_ID=
+# JARVIS_CF_TURN_API_TOKEN=
+# The backend mints short-lived credentials from these (24h TTL, cached) and
+# serves them to both the browser (/api/voice/ice) and its own RTCPeerConnection.
+# Alternative/fallback — static ICE list (used when the CF vars are unset):
+# JARVIS_WEBRTC_ICE=stun:stun.l.google.com:19302,turn:user:pass@turn.example.com:3478
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -74,13 +74,16 @@ CORS_ORIGINS=*
 # JARVIS_REPO_URL=
 
 # ─── WebRTC TURN (voice from CGNAT/cellular networks) ────────────────────────
-# This deployment has no public UDP inbound (web rides a cloudflared tunnel),
-# so WebRTC voice from outside the LAN/tailnet (e.g. iPhone on 5G) needs a
-# TURN relay. Recommended: Cloudflare Realtime TURN — create a TURN app at
-# dash.cloudflare.com → Realtime → TURN, then set BOTH:
+# Needed only when voice is used from OUTSIDE the host's LAN/VPN (e.g. phone
+# on 5G) and the host has no public UDP inbound. PREFERRED setup path: the
+# Settings → Voice → "Cloudflare TURN (voice relay)" section (or the Setup
+# Wizard services step) — stores the key encrypted in the DB, applies live,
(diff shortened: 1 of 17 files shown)


omnigentx:
Added: TURN credentials are now configurable from **Settings → Voice** and the **Setup Wizard** (stored encrypted in the config DB — env vars remain a headless fallback; DB takes priority). Includes OSS-facing walkthrough copy in both surfaces + SELF_HOSTING.md §9. Backend 87 tests / frontend 2 new e2e specs green.

omnigentx (commented):
## Code review — Cloudflare TURN

Strong PR: `get_ice_servers()` is a genuine single source of truth (both `/api/voice/ice` and the aiortc peer read it), no silent fallbacks (mint failure is loudly logged, stale-serve is `warning`), DB-over-env priority is tested, and the suite includes a real aiortc loopback (no mock of the changed subsystem) plus two e2e flows. Two non-blocking findings:

**[F1 · LOW · security/robustness]** `webrtc_voice.py` — `key_id` is interpolated into the mint URL path unescaped:
```python
_CF_TURN_MINT_URL.format(key_id=key_id)
```
`key_id` is user-supplied config so risk is low, but a stray `/` or whitespace silently produces a wrong path + confusing 404 instead of a clear "bad key_id". Consider `urllib.parse.quote(key_id, safe='')` for a crisp failure mode.

**[F2 · INFO · acceptable]** `WebRtcVoiceSession(ice_servers=None)` self-serves via `get_ice_servers()`, which can block ≤10s on a cold CF mint. The ws handler correctly pre-fetches off-loop (`asyncio.to_thread`) and the docstring warns callers — fine as documented. Just ensure no event-loop path ever relies on the `None` default.

omnigentx:
Review addressed in `7deddf2`:
- **F1**: `key_id` is now percent-encoded (`urllib.parse.quote(key_id, safe='')`) before URL interpolation — a stray `/` fails as a clear CF 404 instead of hitting a different API path. Covered by new unit test `test_mint_url_percent_encodes_key_id` (suite: 10 passed).
- **F2**: acknowledged — no event-loop path uses the `None` default; the ws handler pre-fetches via `asyncio.to_thread` and the docstring warns future callers.

PR #84, opened 2026-06-10
fix(auth): reload after modal login so 401-poisoned views recover

omnigentx:
## Bug

After a passkey (or API-key) sign-in via the AuthGate modal, some routes kept showing 401-failed/empty state until the user manually reloaded the page. Previously "fixed", but it kept happening.

## Root cause

`AuthGate` is an **overlay, not a route guard** (`App.vue`): deep-linking to a view while logged out mounts it behind the modal, its `onMounted` fetch 401s, and nothing replays it. The previous fix wired `EVENTS.RESTORED` re-fetch into **ChatView only** (plus the SSE composables) — every other view (~20) and `AppLayout`'s boot fetches stayed 401-poisoned after login.

**Scope note (review F-84.5):** this fixes recovery after a **lockout** login (`unauthenticated → authenticated`, AuthGate modal — passkey or API key, local or cross-tab). A transient mid-session 401 that self-heals via `challenged → authenticated` intentionally does NOT reload (in-page state there is still healthy); that path keeps the pre-existing per-listener wiring (SSE + ChatView). `TeamMonitor` was the worst case: it sits in `<keep-alive>`, so even navigating away/back never remounted it.

## Fix (class-level, not per-view)

- `stores/auth.js`: `RESTORED` now carries `{ from: previousStatus }` (both local transitions and cross-tab `_applyRemoteTransition`).
- `App.vue`: on `RESTORED` with `from === 'unauthenticated'` (modal login recovering a lockout) → `window.location.reload()`. Boot probe (`from === 'unknown'`) and challenged-recovery never reload → no reload loop. Locked sibling tabs recover identically via BroadcastChannel.
- At lockout time there is no in-page state worth preserving — it is all 401-poisoned, so a reload is the cleanest full replay and removes the need to wire (and remember to wire) a RESTORED listener into every current and future view.

## Latent race surfaced & fixed

`main.js` mounted the app before `router.isReady()` → `route.meta` read `{}` for one tick, so the boot auth probe ran on bare `/setup` pages (flipping the store to `unauthenticated` mid-wizard — which made the wizard's inline login trigger the new reload and restart the wizard) and the AppLayout chrome flashed over the wizard. Now mounts after `router.isReady()`.

## Click-flow (e2e: `auth-gate-reload-recovery.spec.ts`)

1. Deep-link `/token-usage` logged out → view mounts behind the gate, metrics fetch 401s
2. Type API key in the modal → app **reloads itself** (spec never calls `page.reload()`)
3. Metrics re-fetched (200) → `1.2M` renders → whoami count goes quiet (no reload loop) → still on `/token-usage`

## Tests

- Unit: 3 new (RESTORED payload from boot probe / modal login / cross-tab) — **156 passed**
- E2E: 1 new happy-path spec above; existing login specs updated for the self-reload via shared `waitForPostLoginReload` harness helper; `whoami` fixtures now two-phase `[false, true]` (sequence support already in harness) — **84 passed**

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -7,7 +7,8 @@ import ConfirmModal from './components/ConfirmModal.vue'
 import AuthGate from './components/AuthGate.vue'
 import { useConfirmState } from './composables/useConfirm'
 import { onSetupRequired, onUnauthorized } from './api'
-import { useAuthStore } from './stores/auth'
+import { useAuthStore, STATUS as AUTH_STATUS } from './stores/auth'
+import { on as onAuthEvent, EVENTS as AUTH_EVENTS } from './auth/bus.js'
 
 // Single globally-mounted confirmation dialog driven by useConfirm().  Views
 // call ``await confirm({...})`` instead of ``window.confirm(...)`` so every
@@ -37,6 +38,21 @@ onMounted(async () => {
   // (re-probes once before locking the UI; see stores/auth.js#on401).
   onUnauthorized((reason) => {
     auth.on401(reason)
+  })
+
+  // Recovering from a LOCKOUT (AuthGate modal login — passkey or API key)
+  // reloads the page. Views mount behind the gate overlay (it is not a
+  // route guard), so their onMounted fetches already 401'd; only SSE
+  // composables and ChatView replay on RESTORED, every other view stays
+  // broken until a refresh. Rather than wiring a RESTORED listener into
+  // 20+ views (and every future one), replay everything: at lockout there
+  // is no in-page state worth preserving — it is all 401-poisoned.
+  // Boot probe (from === 'unknown') and challenged-recovery must NOT
+  // reload — those paths have healthy in-page state.
+  onAuthEvent(AUTH_EVENTS.RESTORED, ({ from } = {}) => {
+    if (from === AUTH_STATUS.UNAUTHENTICATED) {
+      window.location.reload()
+    }
   })
 
   // Boot probe: ask the backend whether the existing cookie is still good.
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -14,4 +14,9 @@ app.use(router)
 // resolves makes meta read as {} for one tick — the probe then runs on
 // /setup, flips the auth store to 'unauthenticated' mid-wizard, and the
 // AppLayout chrome (with its SSE streams) flashes over the wizard.
-router.isReady().then(() => app.mount('#app'))
+// isReady() rejects if a navigation guard throws — degrade to mounting
+// anyway (router lands on the resolved-or-error state) instead of a
+// blank page with no surfaced error.
+router.isReady()
+  .catch((err) => console.error('[main] initial route failed:', err))
+  .then(() => app.mount('#app'))
--- a/frontend/src/main.js
+++ b/frontend/src/main.js
@@ -8,4 +8,10 @@ import './assets/tokens.css'
 const app = createApp(App)
 app.use(createPinia())
 app.use(router)
-app.mount('#app')
+// Wait for the initial route: App.vue's onMounted branches on
+// route.meta.layout (bare /setup pages skip the boot auth probe, and the
+// template picks bare vs AppLayout chrome). Mounting before the router
+// resolves makes meta read as {} for one tick — the probe then runs on
+// /setup, flips the auth store to 'unauthenticated' mid-wizard, and the
+// AppLayout chrome (with its SSE streams) flashes over the wizard.
+router.isReady().then(() => app.mount('#app'))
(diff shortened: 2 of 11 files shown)


omnigentx:
## Code Review — `fix(auth): reload after modal login so 401-poisoned views recover`

**Verdict: Approve with 4 minor/nit findings.** Core logic verified against the auth state machine + App.vue ordering + main.js. No blocking correctness issues.

### What's good (verified)
- **`from` discriminator is the right SSoT.** Traced the state machine: only `unauthenticated → authenticated` (modal login / cross-tab) reloads. `on401 → challenged → probe → setAuthenticated(prevStatus=challenged)` carries `from='challenged'` → **no** reload (in-page state preserved); boot probe carries `from='unknown'` → no reload. There is **no automatic path** that reloads mid-work: from `unauthenticated`, `refresh()` is guarded (`status !== AUTHENTICATED` returns) and `on401` early-returns, so only an explicit user login leaves lockout.
- **Listener ordering is correct.** The `RESTORED` listener (App.vue:52) is registered *before* `auth.init()` (App.vue:62), so the boot probe's `RESTORED{from:'unknown'}` is observed but correctly ignored — no reload loop.
- **e2e respects Happy-Path-FIRST.** New spec mocks no part of the changed subsystem, never calls `page.reload()` (the app self-reloads), and adds a negative guard on the whoami count to catch a reload loop.
- **main.js race fix is accurate.** Both `useBareLayout` (App.vue:24) and the boot-probe skip (App.vue:61) read `route.meta.layout`; mounting before `router.isReady()` makes meta `{}` → the probe runs on `/setup`. Fix is sound.

### Findings

**F-84.1 — Every login now reloads, not just deep-links (UX trade-off, INFO).**
The common path (boot at `/` → boot probe → `unauthenticated` → gate → login) also crosses `unauthenticated → authenticated`, so **every** login does a full `reload()` plus an extra whoami round-trip, and drops any optimistic in-gate state. This is the intended cost of the class-level fix and is acceptable — just confirming it's intended for the common case, not only deep-links. Non-blocking.

**F-84.2 — `router.isReady()` has no `.catch` (robustness, MINOR).**
`main.js`: `router.isReady().then(() => app.mount('#app'))`. `isReady()` resolves on redirects but **rejects if a navigation guard throws** → the app never mounts → blank white screen with no surfaced error. Suggest `.catch(err => { console.error(err); app.mount('#app') })` so a guard error degrades to a working app instead of a blank page. Low likelihood, cheap insurance.

**F-84.3 — `reload()` runs before `_broadcast()` (fragile ordering, NIT).**
`stores/auth.js:306–307`: the synchronous `emit(RESTORED)` (whose App.vue listener calls `window.location.reload()`) precedes `this._broadcast(expiresIn)`. It works only because `reload()` does not interrupt the currently-executing synchronous task, so `_broadcast` still fires. This is correctness-by-luck on `reload()` semantics. Suggest broadcasting **before** emitting `RESTORED` so cross-tab notification is guaranteed independent of the reload listener.

**F-84.4 — Fixture `backend_source` uses wrong `dashboard/` prefix (doc nit).**
`auth_gate_reload_recovery.yaml` references `dashboard/src/App.vue` and `dashboard/src/views/TokenUsage.vue`, but the real tree is `frontend/src/...`. It's documentation metadata only, but it diverges from the actual source layout — please fix the prefix to `frontend/`.

**F-84.5 — Scope is narrower than the description implies (scoping, INFO).**
The PR frames the fix broadly ("every view stays 401-poisoned", "replay everything"), but the reload only covers `unauthenticated → authenticated`. A transient mid-session 401 that recovers via `challenged → authenticated` (`from='challenged'`) does **not** reload, so a non-ChatView view that 401'd during that blip can stay stale (still relying on the old per-view `RESTORED` wiring = SSE + ChatView only). This is a pre-existing gap and scoping to the lockout case is reasonable — just recommend wording the description as "recovers after a *lockout* login" so it isn't read as fixing all recovery paths.

### Tests
The 3 new unit tests (`from` payload for boot probe / modal login / cross-tab) match the implementation exactly and will pass. The new e2e plus the 3 updated specs route through `waitForPostLoginReload` (awaits the post-reload whoami), which is the right deterministic signal. I did not check out the PR branch to run them (the working tree has uncommitted changes) — verification here is by reading, not by running the branch.

omnigentx:
Review addressed in `475a5a9`:
- **F-84.1**: confirmed intended — every lockout login (deep-link or boot-at-`/`) reloads; the extra whoami round-trip is the accepted cost of the class-level replay.
- **F-84.2**: `router.isReady()` now has a `.catch` that logs and mounts anyway — a guard error degrades to a working app, not a blank page.
- **F-84.3**: `_broadcast()` now runs BEFORE the local `RESTORED` emit, so sibling-tab notification never depends on `reload()` letting the current task finish.
- **F-84.4**: fixture `backend_source` prefix fixed to `frontend/`. Note: 39 older fixtures still use the legacy `dashboard/` prefix — left for a separate cleanup to keep this PR scoped.
- **F-84.5**: PR description updated with an explicit scope note — this covers **lockout** recovery only; `challenged → authenticated` intentionally keeps in-page state (no reload).

Verification after the changes: unit 156 passed, e2e 84 passed (full suite, production build).

PR #85, opened 2026-06-11
feat(context): backend-managed context compaction with versioned visibility

omnigentx:
## What

Backend-managed **context compaction with versioned visibility** for long-running agents, per `docs/context-compaction-versioned-visibility-spec.md` (shipped in this PR together with an implementation addendum — §16 — recording every deviation and why).

When an agent's context crosses the configured threshold before an LLM call, the backend:

1. saves an append-only **raw snapshot** (`agent_context_snapshots`, trigger `pre_compaction`) — raw history is never deleted,
2. runs the **rule-based compactor** (same output contract a future LLM compactor must honour),
3. **validates the plan independently** (templates preserved, latest user request verbatim, tool-call/result pairing intact, final message byte-identical, summary sections complete, savings floor),
4. builds the working context from deep copies and applies it in one `load_message_history()` call,
5. records a `context_compaction_events` row (working json + plan + stats) and emits `context_compaction_started/completed/failed` on the existing SSE activity stream — **no polling**.

The agent continues its task seamlessly: the provider payload is rebuilt from `message_history` + delta on every call (`LlmDecorator._prepare_llm_call`), so the imminent call picks up the compacted history; the pending delta is never touched.

### Dashboard

- **Agent Detail → "Context Versions" tab**: compaction timeline (msgs/tokens before→after, saved %, trigger, confidence), failed compactions inline, lazy-loaded summary, lazy-loaded **Before/After diff** with per-message dispositions (kept / truncated / dropped / summary). Live "Compacting…" banner while in flight.
- **Settings → Context Compaction**: enable toggle, threshold ratio, max context tokens (0 = auto from the model's context window), keep-recent, tool-result cap, min savings, versions visible, live-status toggle. PATCH sends only changed keys → clean audit history in `config_history`.
- **Toasts** (bilingual): completion shows tokens saved; failure says the agent continues on raw context.

## Spec critique — what was good, what was fixed

**Kept from the spec (good design):** raw append-only vs working separation; backend-side validation (the agent never rewrites its own history unchecked); rule-based MVP behind the future LLM-compactor contract; SSE-only status; metadata-first lazy API; the mid-flight tool-pairing risk callout.

**Fixed during implementation** (full rationale in spec §16):

| # | Spec said | Shipped | Why |
|---|-----------|---------|-----|
| 1 | Resume: "prefer working snapshot, fallback raw" | **newest-wins across both** | raw snapshots keep landing after a compaction; preferring an older working snapshot would resume into the past and silently drop turns |
| 2 | Two tables (working snapshots + events) | **one `context_compaction_events` table** | strict 1:1, ~8 duplicated columns; derived stats computed at API layer, never stored |
| 3 | "preserve the current delta" | **pair-extended tail cut + final-message byte-identity + forced latest-user keep** | mid tool-loop, history already ends with `assistant(tool_calls)` whose results ride in the delta — protecting only the delta is insufficient |
| 4 | token estimation unspecified | trigger = provider-reported context tokens (fallback chars/4); savings = same estimator both sides | mixing real "before" with estimated "after" fabricates savings |
| 5 | `max_context_tokens: 120000` static | **0 = auto from model context window** | survives model swaps; 120k stays as last-resort fallback |
| 6 | — | per-agent lock + per-history-length memo | prevents compaction storms when a plan keeps failing the savings floor |
| 7 | `GET/PATCH /api/settings/context-compaction` | `/api/context-compaction/settings` | collides with generic `/api/settings/{category}`; still backed by config_service (audit/export free) |
| 8 | hex palette (`#0a0d14`…) | `tokens.css` variables | spec's hexes don't exist in the design system; hardcoded hex is a repo anti-pattern |

**Scope cuts (explicit, not silent):** hook attaches to in-process agents only (chat path — same wiring as the always-on token hook); subprocess team agents benefit via the resume path. `promote_to_memory` stays in the contract but is always empty (no memory subsystem exists). `tool_result_artifacts` table deferred (raw_references point at snapshot + message range).

## Tests

**Backend (30 new, all green):**
- `tests/test_services/test_context_compaction.py` (21) — real `PromptMessageExtended` + real `prompt_serialization` + real SQLite (no mocks of the subsystem under test): planner zones, pair-safe cut sweep (keep_recent 2–8), forced latest-user keep, round-trip + full validator matrix (dropped user / broken pairing / missing sections / template removal / savings floor / manual bypass), oversized tool-result truncation with raw ref, **full pipeline** (raw snapshot byte-identical, completed event row, live history swapped), failure leaves history untouched + failed row recorded, memo guard, hook never raises, **resume newest-wins both directions**, failed events ignored on resume.
- `tests/test_routes/test_context_compaction_routes.py` (9) — real config_service + real snapshot DB: settings defaults/PATCH persistence/422/auth, versions list (default limit follows the `snapshot_versions_visible` setting), detail scoping, empty state, diff dispositions, failed-event 404.

**Frontend:**
- Unit (node:test): 4 new store cases (started/completed/failed + pause-cycle no-clobber) — **160/160 pass**.
- Playwright e2e (6 new, real click-flows, no mocking of the changed subsystem): versions timeline + lazy summary + lazy diff (with request-count assertions), SSE completed → toast, SSE started → live banner, empty state, settings load → changed-keys-only PATCH → success msg, 422 surfaces inline. **Full e2e suite: 90/90 pass.**

**Known coverage gap (stated, not hidden):** the harness delivers mock SSE bodies atomically, so the "refetch versions list when a compaction completes while the tab is open" watch isn't observable e2e; its store-side inputs are unit-covered and the fetch path is the same one the tab-open flow exercises. A real-backend e2e of an actual threshold-triggered compaction during chat requires a live LLM loop — exercised by the service-layer full-pipeline test against the real agent-history API instead.

**Note on AgentDetail/Settings copy:** the new Context Versions tab and SettingsCompaction page follow the existing convention of their surfaces (English-only, like every other AgentDetail tab and settings page — review F8). The AppLayout toasts are bilingual per `useLang()`. A bilingual sweep of AgentDetail is pre-existing scope, flagged separately.


omnigentx:
# Adversarial review — PR #85 (context compaction with versioned visibility)

**Verdict: Request changes** — no blockers, but two MAJOR findings (a user-visible stuck-state bug and a silent feature gap on dynamic agent reload) should be fixed before merge. The core safety design (pair-extended tail cut, final-message byte-identity, deep-copy build, validate-before-apply, newest-wins resume) held up under scrutiny; I verified the load-bearing claims in source rather than trusting the PR body, and re-ran all test suites myself.

## Findings table

| ID | Severity | Where | One-liner |
|----|----------|-------|-----------|
| F1 | MAJOR | `context_compaction.py:752, 785-787` + `agents.js` started case | `context_compaction_started` emitted before planning; plan-None and unexpected-exception paths emit no terminal event → "Compacting context…" banner stuck forever |
| F2 | MAJOR | `dynamic_agents.py:142-152` | Dynamic-agent reload re-attaches only the token hook — agents registered/reloaded after startup silently never compact |
| F3 | MINOR | `agent_spawner_server.py:650` | Second resume path (`resume_spawn`) still raw-only; resume rule now split across two readers; PR-body claim overstated |
| F4 | MINOR | `routes/settings.py:244` vs `context_compaction.py:164` | Range validation bypassable via generic `PUT /api/settings/{category}/{key}`; `_coerce` silently degrades garbage to defaults |
| F5 | MINOR | `context_compaction.py:142-143` | `get_compaction_config` fails open (enabled=True defaults) when the config DB is unreadable, DEBUG-level log only |
| F6 | MINOR | `AgentDetail.vue:200-213, 236-253` | Versions fetch error renders as the "No context compactions yet" empty state; detail/diff errors cached with no retry |
| F7 | MINOR | `context_compaction.py:381-393` vs `518-528` | Plan flags truncation by joined text length but `_truncate_tool_result` truncates only per-block → "truncated" disposition can be a no-op lie in the diff UI |
| F8 | MINOR | `SettingsCompaction.vue`, `AgentDetail.vue` versions tab | English-only user-visible copy vs the repo bilingual rule (AgentDetail acknowledged in PR body; SettingsCompaction not) |
| F9 | NIT | `context_compaction.py:637-638, 864` | Lock/memo keyed on normalized name — `[1]`/`[2]` instances share guards; dicts never pruned |
| F10 | NIT | `context_compaction.py:323-343, 736+`; diff route | Whole pipeline runs synchronously on the event loop; `_pair_safe_tail_start` O(n²) worst case; diff endpoint parses full snapshots inside an `async def` |
| F11 | NIT | tests | No test pushes a compacted history (summary-user adjacent to user messages) through a real provider converter |
| F12 | NIT | `context_compaction.py:401-405, 527, 707` | `raw_references` range includes the force-kept user index; `reason="manual"` has no caller; f-string without placeholder |
| F13 | NIT | spec §8 vs `validate_working_context` | "remove pinned messages" validator rule neither implemented nor recorded as a deviation in §16 |

## Detailed findings

### F1 (MAJOR) — started SSE without a terminal event → stuck "Compacting…" banner
`_run_compaction` emits `context_compaction_started` (context_compaction.py:752) **before** planning. If `plan_compaction` returns `None` (middle zone too small, line 785-787), the function returns with no `completed`/`failed` event and no event row. The frontend store sets `compaction.inProgress = true` on `started` and clears it **only** on `completed`/`failed` (agents.js, new cases) — the AgentDetail banner ("Compacting context…", AgentDetail.vue:1167-1170) then shows indefinitely.

Reachable scenario: default `keep_recent_messages=10`, 2 template messages, 13-14 message history with huge tool results exceeding 0.7× the window. `maybe_compact_agent`'s length gate (`len >= keep_recent + 3`, line 704) passes, but `tail_start - head_end = len - 10 - 2 < 3` → plan None. The memo (line 725) blocks retry until the history grows, then the same thing repeats: each round emits another orphan `started` **and writes another full `pre_compaction` raw snapshot** (line 776 runs before planning) — UI stuck + snapshot bloat.

Same leak on the unexpected-exception path: anything raised inside `_run_compaction` outside a `_fail` wrapper (e.g. `to_json(working)` at step 6, line 809-814 — *after* the live history was already swapped) is swallowed by `maybe_compact_agent`'s outer catch (line 731-733) with no terminal SSE and no event row.

**Fix:** run the cheap planning before emitting `started` (or emit a terminal "skipped"/`failed` status on the plan-None path), and wrap the whole of `_run_compaction` so any escape still emits `context_compaction_failed`. Also consider not saving the raw snapshot until a plan exists. Add a test: threshold exceeded + middle-too-small → no orphan `started`, no stuck `inProgress`.

### F2 (MAJOR) — dynamic agent reload silently drops the compaction hook
`server.py:489-493/514-517` attaches the hook at startup and after preload. But the dynamic-reload poll loop (`services/dynamic_agents.py:142-152`) re-attaches **only** `attach_token_persistence_hooks_to_all` after `load_agent_data` — exactly the path whose comment explains why newly registered agents need always-on hooks re-attached. Any agent created or reloaded through the dynamic path after startup gets token tracking but **no compaction**, with zero log signal. The feature degrades silently for precisely the long-running dynamic agents that need it most.

**Fix:** add `attach_compaction_hooks_to_all(agent_app)` next to the token-hook re-attach in `dynamic_agents.py` (it is already idempotent via the `_jarvis_compaction_hook` sentinel). Better long-term (item 10 of the repo checklist): one `attach_always_on_hooks(agent_app)` helper so the next always-on hook is attached "for free" instead of requiring three call sites to be kept in sync — this PR just demonstrated how that duplication fails.

### F3 (MINOR) — resume rule split across two readers
`inject_resume._load_context_from_db` now uses `load_latest_context_json_any` (newest-wins) — but the MCP resume path `resume_spawn` (`fast-agent/src/fast_agent/spawn/servers/agent_spawner_server.py:650-651`) still calls raw-only `load_latest_context_json`, under a comment declaring `agent_context_snapshots` the "single source of truth" for resumable history. Today this can't diverge (the hook only covers in-process agents, which `resume_spawn` never resumes), but the §16.9 roadmap explicitly extends compaction to subprocess agents — at that point the two resume paths return different histories for the same agent, the classic two-readers SSoT violation the repo guidelines call out. The PR body's "subprocess team agents benefit via the resume path" is also only true for the inject path. **Fix:** point `resume_spawn` at `load_latest_context_json_any` too (or document in §16 why it intentionally stays raw-only and update the now-overbroad SSoT comment).

On the newest-wins rule itself: I looked for a data-loss scenario and found none in-process — post-compaction `chat_complete` raw snapshots are themselves based on the compacted history, a crash between event-write and the next raw snapshot resumes from the working snapshot (correct), and a `created_at` tie resolves to the raw row (`max()` returns the first maximal candidate, context_persistence.py:795) which is the safe direction. Cross-process clock skew is shared with every other `created_at` ordering in the app — not new risk from this PR.

### F4 (MINOR) — settings validation has a second, unvalidated writer
The new typed endpoint enforces ranges via `validate_config_updates`, but the pre-existing generic `PUT /api/settings/{category}/{key}` (`routes/settings.py:244`) writes the same `context_compaction` rows through `config_service.set` with no range check. `compact_at_ratio=0.01` (compaction storm) or `keep_recent_messages=0` go straight in; a non-numeric value is silently coerced back to the default at read time (`_coerce`, context_compaction.py:110-123) with no warning. One state, two write paths, one validated. **Fix:** enforce at the single choke point all writers share — clamp/validate in `get_compaction_config` on read (log loudly when a stored value is out of range), which also covers hand-edited DBs.

### F5 (MINOR) — fail-open on config read failure
`get_compaction_config` swallows any config-DB failure and returns `DEFAULTS` — which includes `enabled: True` (context_compaction.py:142-143, DEBUG log). A user who disabled compaction gets it re-enabled by a transient config-DB error, for a feature that rewrites live agent history. Fail-closed (treat unreadable config as `enabled=False`) is the safer direction here, and the failure deserves at least WARNING.

### F6 (MINOR) — UI error states
- `fetchContextVersions` catch (AgentDetail.vue:209-211) only `console.error`s; the template then falls through to the "No context compactions yet" empty state — an API failure masquerades as ground truth ("error messages actionable" checklist item). Show a distinct error state with retry.
- `versionDetails[v.id] = { error: true }` / same for diffs (AgentDetail.vue:236-253) is cached permanently — after one failed lazy-load, re-clicking never refetches; the only recovery is a page reload. Clear the cache entry on error or offer a retry button.

### F7 (MINOR) — truncation flag vs truncation action mismatch
`plan_compaction` flags a tool result when the **joined** text of all blocks exceeds `max_chars` (context_compaction.py:383-393), but `_truncate_tool_result` only shortens **individual blocks** longer than `max_chars` (518-528). A result composed of many medium blocks is flagged `truncate_tool_result`, the diff UI renders it "truncated", yet nothing was removed. The stored plan and the actual working context disagree — the exact "UI shows what the applied plan did" guarantee the diff endpoint's docstring promises. **Fix:** truncate against a per-result budget across blocks, or flag per-block.

### F8 (MINOR) — bilingual rule
`SettingsCompaction.vue` is entirely English (intro, labels, hints, "Save Changes", "Saved · live now"), as is the new versions tab. Repo rule: "user-visible UI copy → bilingual via `useLang()`; single-language literals are bugs." I checked the convention claim: of the settings pages only `SettingsVoice.vue` uses `useLang`, so this matches the surrounding (pre-existing) debt, and the AppLayout toasts are properly bilingual. Acceptable as consistent-with-convention, but the PR body only flags AgentDetail — SettingsCompaction adds ~15 new English-only strings to the backlog and should be named in that flagged scope too.

### F9-F13 (NITs)
- **F9:** `_locks`/`_last_attempt_len` are keyed on `normalize_agent_name` output — `FinanceAgent[1]` and `[2]` share a memo, so one instance's attempt at length L suppresses the other's at the same length. Dicts are never pruned (bounded by agent-name count, so mostly cosmetic).
- **F10:** the whole pipeline (raw `to_json`, `estimate_tokens` via per-message `model_dump_json` twice, round-trip validation, deep copies, SQLite writes) runs synchronously inside `before_llm_call` — for a 5 MB context this stalls the event loop (SSE, other requests) noticeably; `_pair_safe_tail_start` is O(n²) in the pathological case. Verified the flip side: because nothing in the critical region yields, in-process races are impossible and the lock is belt-and-braces — but a `to_thread` offload (with the lock kept) would be kinder to the loop.
- **F11:** no test converts a compacted history through a real provider converter. The summary is `role="user"` inserted directly before kept user messages (user,user adjacency); `AnthropicConverter.convert_to_anthropic` maps 1:1 with no alternation merging (llm_anthropic.py:2397). Likely fine for current provider APIs, but a one-line converter test would pin it.
- **F12:** `raw_references.message_indexes = [middle[0], middle[-1]]` after the forced user index was removed from `middle` — the stored range claims the kept message was summarized. `reason="manual"` plumbing (incl. validator bypass) has no caller in this PR. `f"\n…[truncated…"` (line 527) has no placeholder.
- **F13:** spec §8 "reject plans that remove pinned messages" — no pin concept exists in the codebase and the validator doesn't implement it, but §16 doesn't record the cut. Add one line to the deviation list (it's otherwise commendably explicit).

## Verified claims (checked in source, not trusted)

- Mid-loop pairing safety: confirmed `before_llm_call` fires before each loop iteration (tool_runner.py:336-341) and the provider payload is rebuilt from `_message_history` + delta each call (llm_decorator.py:873-876); the pair-extended cut + final-message byte-identity check make the claimed invariant real. `load_message_history` deep-copies (llm_decorator.py:1316-1321).
- Validate-before-apply: live history is only mutated at step 5 after the violation list is empty; the failed-validation test asserts byte-identical history afterwards.
- Concurrency: every statement between history capture and lock acquisition is synchronous, so the feared "stale-history compaction clobbers a concurrent turn" interleaving is unreachable in single-threaded asyncio. Re-compaction loops are bounded by the memo; a previous summary always lands in the middle zone whenever a plan exists, so the "exactly 1 summary" validator can't deadlock repeated compactions.
- Auth: all five new endpoints carry `Depends(verify_api_key)`; diff payload is bounded to 200-char previews per message.
- Tests are genuinely integration-grade: real `PromptMessageExtended`, real `prompt_serialization`, real SQLite via `SPAWN_REGISTRY_DB`; route tests run the real `config_service`. The only pinned piece is the config reader (reasonable). No mock-returns-post-mutation-state pattern found.

## Test verification (run by reviewer)

| Suite | Result |
|---|---|
| `backend: tests/test_services/test_context_compaction.py + tests/test_routes/test_context_compaction_routes.py` | **30 passed** (21 + 9) |
| `backend: tests/test_context_persistence.py` (pre-existing, touched module) | **32 passed** |
| `backend: tests/test_services/test_inject_resume.py` (changed module) | **9 passed** |
| `frontend: npm run test:unit` | **160 pass / 0 fail** |
| `frontend: npx playwright test agent-context-versions.spec.ts settings-compaction.spec.ts` (after fresh build) | **6 passed** |

Coverage gaps worth closing with the F1/F2 fixes: the plan-None-after-started path (F1) and dynamic-reload hook attachment (F2) have no tests — both are exactly the kind of silent-degradation paths the existing matrix otherwise covers well.

omnigentx:
Thanks for the review — both MAJORs were real. All fixes in `2255e96`, re-verified green (backend affected suites **1306 passed**, frontend unit **160**, e2e **91**).

| ID | Status | Resolution |
|----|--------|------------|
| F1 | ✅ Fixed | Feasibility pre-check (deterministic planner pass with a placeholder id) runs **before** `started` is emitted or the raw snapshot is saved — an infeasible plan is now a total no-op: no stuck banner, no orphan snapshot per LLM call. Once `started` is out, a blanket `except` inside the pipeline guarantees a terminal `failed` event. Pinned by `test_infeasible_plan_is_silent_noop` (asserts zero SSE events + zero snapshot rows) and `test_unexpected_pipeline_error_emits_terminal_failed` (asserts `started`→`failed` sequence + untouched history). |
| F2 | ✅ Fixed | `dynamic_agents.py` reload loop now re-attaches the compaction hook next to the token hook (same idempotent sentinel). Helper merge/idempotency pinned by `test_attach_compaction_hooks_idempotent_and_merges`. |
| F3 | 🟨 Acknowledged, deferred | `resume_spawn` raw-only is intentionally untouched: subprocess agents are explicitly out of compaction scope in v1 (addendum §16.9), so no working snapshot can exist for them today — newest-wins and raw-only are equivalent there. Agreed it must flip the day subprocess compaction lands; that change belongs to the same PR that extends scope, where it can be tested against a real working snapshot. |
| F4 | ✅ Fixed | Ranges single-sourced in `_RANGES` and **re-enforced at read time**: rows written through the generic settings PUT that are out-of-range or unparseable revert to the key's default with a WARNING. `test_out_of_range_config_value_reverts_to_default`. |
| F5 | ✅ Fixed | Config-DB read failure now fails **closed** (compaction disabled for that call, WARNING log). `test_config_fails_closed_when_db_unreadable`. |
| F6 | ✅ Fixed | List fetch error renders a distinct error state with Retry (no longer masquerades as the empty state); lazy detail/diff errors are never cached and have explicit Retry buttons. New e2e: `versions fetch error shows error state, Retry recovers`. |
| F7 | ✅ Fixed | Truncation flagging is now per-block, matching exactly what `_truncate_tool_result` cuts. `test_truncation_flag_is_per_block_not_joined` (5×1500-char blocks, joined 7500 > cap → NOT flagged). |
| F8 | 🟨 Acknowledged | Left English-only deliberately to match every existing settings page (same debt class as AgentDetail); now flagged in the PR body alongside it. A bilingual sweep of settings + agent-detail is a separate, mechanical task — happy to file it. |
| F9 | 🟨 By design | Locks/memos are per-process, and so is everything they guard: `agent_app._agents` (the live histories being compacted) is per-process state. A multi-worker deployment would break the agent runtime itself long before the compaction guards. |
| F10 | 🟨 Accepted | The pipeline is CPU-bound but bounded: the cut sweep is O(keep_recent × tail) with keep_recent ≤ 100, and compaction itself keeps history length bounded. Will profile if real-world stalls show up; moving to a thread would forfeit the "no yielding awaits in the critical region" property your review verified. |
| F11 | 🟨 Residual risk, documented | Verified in source: the Anthropic converter maps messages 1:1 (no role merging), and the Anthropic Messages API documents that consecutive same-role messages are combined into one turn; OpenAI likewise accepts them. Strict-alternation providers (Gemini) remain a residual risk — if it materializes, the fix belongs in the provider converter layer (where `_merge_trailing_tool_result_turn` already lives), not in compaction. Will smoke-test against a real provider on dev before release. |
| F12 | 🟨 Accepted | `raw_references` gives the middle's index *range*; with a forced-keep inside it the range is approximate by one message. The plan's `keep_verbatim`/`delete_from_working_context` lists are the exact record; the range is a human hint in the summary text. The `manual` reason plumbing is forward wiring for the spec's manual-compaction trigger — kept because the validator's savings-floor bypass for it is contract, not dead code. |
| F13 | ✅ Fixed | Spec §8's "pinned messages" validation cut is now stated explicitly in §16.9 (no pinning subsystem exists anywhere in the repo to source pins from) instead of being silent. |

The stuck-banner scenario (F1) and the dynamic-reload gap (F2) are both now covered by tests that fail on revert.

omnigentx:
# Re-review — PR #85 fix verification (commits `2255e96`, `795a7e4`)

**Verdict: Approve.** All seven "Fixed" findings verified in source against their original failure scenarios; both deferral rationales hold; all suites re-run green by reviewer.

## Fixed findings

| ID | Verdict | Evidence |
|----|---------|----------|
| F1 | **VERIFIED** | Feasibility pre-check (`context_compaction.py:793`) runs before `started`/snapshot; the planner is pure (no IO/emission — checked `plan_compaction` + helpers), and `maybe_compact_agent` memoizes attempt length *before* `_run_compaction` (line 761), so the infeasible case is a true no-op per history length, not per LLM call. Post-`started`, steps 1–5 are inside the blanket `try` → `_fail`. I probed the two remaining escape candidates: (a) `_fail` itself cannot throw — `save_compaction_event` is contractually never-raises (`context_persistence.py:548-582`, catch-all returns None) and `_emit_status` swallows internally; (b) step 6 (`to_json(working)` + completed event) sits *outside* the try, but cannot newly raise: step 4's validator already round-tripped `to_json(working_messages)` on the same object, `load_message_history` deep-copies so `working` is unmutated, and an event-write failure degrades to `event_id=None` with `completed` still emitted. Tests assert zero SSE + zero snapshot rows (infeasible) and `started→failed` + byte-identical history (pipeline error). |
| F2 | **VERIFIED** | `dynamic_agents.py:157-167` — attach is inside the reload **success** branch (after `load_agent_data` returns, before `last_rev = current`), own try/except so a hook failure can't kill the poll loop, idempotent via the `_jarvis_compaction_hook` sentinel. Merge/idempotency pinned by `test_attach_compaction_hooks_idempotent_and_merges` (existing `after_llm_call` preserved, second attach returns 0). The suggested shared `attach_always_on_hooks` helper was not taken — two parallel attach blocks now live in the reload loop; fine for now, worth folding when a third always-on hook appears. |
| F4 | **VERIFIED** | `_RANGES` single source (`context_compaction.py:114-123`), consumed by both `validate_config_updates` and read-time `_in_range`; out-of-range and unparseable rows revert to the key's default with WARNING. Covers the generic PUT bypass and hand-edited DBs at the one choke point all readers share. |
| F5 | **VERIFIED** | Config-DB exception → `enabled=False` + WARNING (`get_compaction_config`). Settings GET UX unaffected: fail-closed triggers only on *exception*, not on missing rows, so a fresh install still shows defaults; when the DB is genuinely unreadable the GET shows disabled — which is the truthful effective state. One residual nit, not a blocker: `reason="manual"` skips the `cfg.enabled` gate in `maybe_compact_agent`, so a future manual trigger would run on guessed defaults under DB failure — worth a guard in the PR that wires the manual caller (F12 plumbing). |
| F6 | **VERIFIED** | `versionsError` renders a distinct error state with Retry (`AgentDetail.vue`), `{error:true}` cache entries are deleted before refetch on re-expand, explicit Retry buttons on detail/diff. New e2e (`versions fetch error shows error state, Retry recovers`) exercises 500→Retry→fixture through the real built UI and asserts the empty state is NOT shown. |
| F7 | **VERIFIED** | Flag condition (`plan_compaction:421`, per-block `len > max_chars`) is now identical to the cut condition (`_truncate_tool_result:560`). `test_truncation_flag_is_per_block_not_joined` pins the original lie scenario (5×1500 blocks, cap 2000 → not flagged). |
| F13 | **VERIFIED** | §16.9 now states the pinned-messages validation cut explicitly. |

## Deferrals (F3, F8–F12)

All reasonable, none a dodge:
- **F3**: sound — the hook covers only in-process agents, `resume_spawn` resumes only subprocess spawns, so no working snapshot can exist on that path today; raw-only ≡ newest-wins. Flipping it belongs to the subprocess-compaction PR where it's testable. Soft note: the original finding also asked to record this in §16 / fix the overbroad SSoT comment at `agent_spawner_server.py:650` — the rationale currently lives only in this comment thread; carry it into the scope-extension PR.
- **F8**: matches the surrounding settings-page convention; now flagged in the PR body. - **F9**: correct — everything the guards protect (`agent_app._agents`) is per-process too. - **F10**: keeping the no-yield critical region over a thread offload is a defensible tradeoff; bounded work. - **F11**: converter-layer ownership is the right call; Anthropic/OpenAI same-role tolerance checked. - **F12**: range-as-human-hint with `keep_verbatim`/`delete` as exact record is acceptable.

## Tests (run by reviewer)

| Suite | Result |
|---|---|
| `backend: test_context_compaction.py + test_context_compaction_routes.py` | **36 passed** (was 30) |
| `frontend: playwright agent-context-versions + settings-compaction` | **7 passed** (was 6) |

Both previously-untested silent-degradation paths (F1 plan-None, F2 hook attach) now have tests that fail on revert. Good to merge.

PR #88, opened 2026-06-11
fix(voice): stop TTS dropping the head of replies + auto-recover WebRTC on mobile

omnigentx:
## What

Two independent voice/WebRTC reliability fixes that surfaced during manual testing.

### 1. TTS dropped the first seconds of replies
\`TtsAudioTrack._next_frame_bytes\` awaited the queue with a 20 ms timeout **in series with** \`recv()\`'s pacing sleep, so every idle frame ran ~1 ms over cadence. That drift (~50 ms/idle-second) was repaid as a faster-than-real-time burst when the next reply arrived — and the client's jitter buffer discards a catch-up burst as "late packets". Result: the reply's head was audibly missing on prod.

- \`_next_frame_bytes\` is now **non-blocking**: it drains only already-arrived data and flushes a partial frame padded with silence (so \`pending()\` can't strand \`wait_drained()\`).
- All pacing lives in \`recv()\`, which **re-anchors its clock to "now"** once lag exceeds two frames (40 ms) instead of bursting.

### 2. WebRTC died on mobile networks
Mobile networks routinely kill the ICE candidate pair mid-session (5G handover, CGNAT rebind, brief radio loss). The old handler treated the first \`failed\` as fatal — error banner, dead mic — even though the \`/ws/voice\` control socket usually survives and can carry a fresh offer.

New \`webrtcRecovery.js\` encodes the re-negotiation policy:
- \`disconnected\` → grace window before forcing a re-offer (cancelled if it recovers on its own);
- \`failed\` → re-offer immediately over the live control WS;
- \`connected\` → reset the retry budget (only *consecutive* failures count);
- past \`maxRetries\` → give up via the existing error banner.

## Tests
- Backend: \`test_webrtc_voice.py\`, \`test_ws_voice.py\` green.
- Frontend unit: \`webrtcRecovery.test.js\` green.
- E2E: \`voice-webrtc.spec.ts\` (mic toggle → valid offer; ICE failure mid-session → re-offer; exhausted retries → banner) green.

### DIFF
--- a/backend/services/webrtc_voice.py
+++ b/backend/services/webrtc_voice.py
@@ -205,6 +205,12 @@ TTS_RATE = 24000      # RealtimeTTS / Edge PCM rate (mono s16)
 WEBRTC_RATE = 48000   # Opus / WebRTC standard
 FRAME_SAMPLES = 960   # 20 ms @ 48 kHz — one outbound frame
 _SILENCE_48K = b"\x00\x00" * FRAME_SAMPLES
+# recv() must never repay pacing debt by emitting frames faster than real
+# time: the receiver's jitter buffer treats a catch-up burst as late packets
+# and DISCARDS them (the "TTS loses its first seconds on prod" bug). Past this
+# lag we re-anchor the pacing clock to "now" instead of bursting. Two frames
+# (40 ms) of slack still lets normal scheduling jitter self-correct.
+_MAX_PACING_LAG_S = 2 * FRAME_SAMPLES / WEBRTC_RATE
 
 FeedAudio = Callable[[bytes], None]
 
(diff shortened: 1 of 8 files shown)


omnigentx:
## Review: PR #88 — voice TTS head-drop + WebRTC mobile recovery

**Risk: LOW–MEDIUM.** Tightly scoped, well-tested. No d=1 breakage. Verified locally:
3/3 new backend tests pass, 7/7 `webrtcRecovery` unit tests pass (e2e spec not run headless).

### What's solid
- **Real hang bug fixed.** The old `_next_frame_bytes` retained a non-frame-aligned
  partial tail across `recv()` ticks on timeout, so `pending()` stayed `True` forever →
  `wait_drained()` hangs → `bot_speaking` stuck high. The new partial-flush (padded with
  silence + buffer clear) fixes it. Covered by `test_tts_track_flushes_partial_tail_*`.
- **Pacing re-anchor is correct.** `_MAX_PACING_LAG_S` (40 ms) re-anchors `_start` instead
  of repaying debt as a faster-than-real-time burst the jitter buffer would drop. The
  `test_tts_track_paces_backlog_instead_of_bursting` guard passes (1.0 s media ≈ 1.0 s wall).
- **Frontend recovery is well-built.** Stale-PC guard (`peer !== pc.value`), consecutive-
  failure budget that resets on `connected`, grace window for `disconnected`, server-error
  path that disables recovery (won't retry a misconfig), and iOS audio-element reuse for the
  autoplay blessing. Backend already closes the prior session on a repeat `webrtc_offer`
  (`ws_voice.py:942`), so the re-offer recovery rides a pre-existing, correct path — no caller
  breakage. `_next_frame_bytes` async→sync change has exactly one caller (`recv()`), updated
  in the same diff.

### Non-blocking notes
1. **Mid-reply underrun now inserts silence instead of waiting (MEDIUM, confirm).** The old
   code held a partial buffer across ticks and *waited* for the next chunk — it never split a
   frame. The new code, on a transient queue-empty *mid-reply*, flushes the partial padded
   with silence and clears the buffer, inserting a sub-20 ms gap and time-shifting the
   following audio (data isn't lost, just delayed). For TTS that streams ahead of real time
   this should be rare, and the code comment acknowledges it — but it's a genuine behavior
   change from "brief wait" to "immediate silence". Worth confirming RealtimeTTS chunking
   doesn't routinely underrun sub-frame during a reply.
2. **No transient "reconnecting" UI state.** Recovery is intentionally invisible, but if a
   reconnect runs the full grace + renegotiation (~4 s+), the user sees the last status frozen
   with no signal. Acceptable; flagging for product call.
3. **Flapping connection has no lifetime cap.** `connected` resets the budget each time, so a
   connection that oscillates connected↔failed can reconnect indefinitely. Edge case, fine to
   leave — a debug log on each reset would help future diagnosis.
4. iOS audio-element-reuse ("reconnected but silent" fix) has no automated coverage —
   acknowledged as un-testable headless. OK.

### Recommendation
**APPROVE** — ship it; treat note 1 as a quick follow-up confirmation, not a blocker.

PR #90, opened 2026-06-14
feat(memory): v2 — LLM-driven recall/capture + GraphRAG on LadybugDB

omnigentx:
Replaces the brittle embedding-prototype intent gate (**topic ≠ intent** — a question like "what's my job?" scored near the fact prototype → stored as memory + no recall) with an **LLM-driven** memory system. SQLite stays the source of truth; **LadybugDB** (embedded graph + HNSW vector) is the rebuildable projection.

## 6 phases
| Phase | What |
|---|---|
| **Recall** | always-retrieve (no gate) + tail-append + change-gate (restart-safe via a `jarvis:recall_key` channel) + provenance sentinel `⟦memory:recalled⟧` + `is_injected_memory()` |
| **Capture** | fast-lane LLM extractor (debounce N turns — a frequency gate, not a content classifier) + **ADD-only** persist (drop write-time curator; conflicting facts kept dated, resolved at read by recency) |
| **Token** | extractor LLM calls tagged `memory:fast` / `memory:slow` → filterable in Token usage |
| **Store** | `vector_backend` flag (LadybugDB default, Qdrant fallback); worker + orchestrator swap; boot migration rebuilds the graph from SQLite; FTS5 stays as the BM25/degraded lane |
| **GraphRAG** | entities flow extractor → `MemoryRecord.entities_json` → worker → `MENTIONS` edges; retrieval does dense + **vector→graph multi-hop** |
| **Slow-lane** | synthesis extractor (workflow/procedural/episodic) piggybacks compaction at the threshold (composite trigger) |

## Also in this PR
- Bulk approve/reject memory candidates (+ UI, toast, loading)
- memory candidates → unified **Approvals inbox** (`pause=False`, two-way SSoT sync)
- `gettime` now returns the weekday
- AgentMemoryPanel design-system pass
- Barge-in context-loss **reproduction** (xfail; root cause = `session_service` cancel-rollback drops the user message)

## Testing
~230 tests. True **e2e** with `PassthroughLLM` playback for LLM calls + deterministic fake embeddings:
- full pipeline (capture → index → LadybugDB → retrieve + entity multi-hop)
- recall through the real orchestrator (LadybugProvider + FTS + RRF + owner-isolation)
- token attribution, migration (rebuild-from-SQLite)

Test isolation: fast-agent sessions (`ENVIRONMENT_DIR`) + LadybugDB redirected to temp dirs so tests never touch the live store.

## Decisions (see `docs/memory-v2-plan.md`)
ADD-only + temporal read-ranking (mem0 2026) instead of curator supersede; LadybugDB (`ladybug` pip pkg, Kùzu-lineage) as the unified graph+vector store, chosen over keeping Qdrant for the native vector→graph fusion.

---

## Review follow-ups (rounds 1–2)

Addressed the ultrareview's gating items + the deferred set:

**Correctness:** symmetric Ladybug/FTS degrade (writer≠reader SSoT fix) · extractor builds its own LLM instance (no shared `usage_accumulator` deflating compaction) · recency/`apply_policy` runs on the happy path + newer-wins tiebreak · slow-lane reads the to-be-summarized segment · memory-loss failures log at WARNING · `degraded_reason` reflects the real backend · exact-dedup keyed on content+scope (not memory_type) · real `token_usage.category` column (memory:fast/slow + compaction).

**Cleanup:** deleted `intent_gate`/`curator`/`conflict` write-path + prototypes + candidate conflict branch + unused `RELATES` table (−726 LOC) · pulled the `useLang` default vi→en flip out (separate sign-off) · i18n'd the last SettingsMemory strings.

**Deferred (noted on the thread):** per-session fast-lane cursor (#5), force-reindex on a real Qdrant→Ladybug switch (#9), cosmetic docstrings + `memory.js` dead state.


omnigentx:
## Review: Memory v2 — LLM-driven recall/capture + GraphRAG on LadybugDB

Reviewed the `feat(memory)` commit (`f6d2f82`, ~50 source + ~33 test files) against `docs/memory-v2-plan.md` (the locked v2 design; note `docs/memory-impl/*` is the *older* Qdrant-era spec that v2 supersedes). I ran the core e2e subset locally: **52 passed** (LadybugDB installed, nothing skipped).

### What's solid

- **Architecture is sound.** SQLite stays the source of truth; the transactional outbox + lease/idempotency worker is intact — only the index target changes. Both Ladybug and Qdrant backends are genuinely wired on read *and* write.
- **Tests are honest and genuinely e2e.** Only the LLM (PassthroughLLM playback) and embeddings (deterministic fakes) are stubbed; the subsystem under test runs through production code (MemoryService → outbox → real worker → LadybugStore/FTS → orchestrator/fusion). No autouse fixture stubs the subsystem under test.
- No silent zero-vector embedding fallback in production (`NullEmbeddingProvider.is_available()==False` and raises; callers gate).
- The Kùzu "no SET on an indexed property" trap is correctly avoided (`upsert_memory` does DETACH DELETE + CREATE in a txn, never `SET emb`).
- Feature flag `memory.enabled` defaults `False` and gates the recall hook / extractor / RPC / boot migration.
- Provenance (sentinel `⟦memory:recalled⟧` + `is_injected_memory()` + channel), tail-append, and the restart-safe change-gate match spec §3. Owner identity is never LLM-supplied; secrets always force approval.
- The barge-in xfail is real (`strict=True`) with a concrete, unfixed root cause: `session_service.py:710-711` pops both the empty assistant *and* the user message on cancel-rollback.

---

### 🔴 Must fix (correctness; some affect already-shipped features)

**1. Asymmetric backend selection writer vs reader → store divergence → silent empty recall.**
`memory_index_worker.py:66-74` catches *any* Ladybug error and silently falls back to Qdrant (WARNING only, no `retrieval_degraded` event). But `orchestrator.py:55-65` on a Ladybug error sets `store=None` → **FTS-only, no Qdrant fallback**. Under a one-sided transient Ladybug failure, writes land in Qdrant while reads hit an empty Ladybug/FTS → **recall returns nothing with no surfaced signal**. This is exactly the writer-path ≠ reader-path SSoT class the repo guidelines warn about. Fix: one authoritative backend decision, symmetric fallback, and surface a degraded event.

**2. The extractor shares `usage_accumulator` with the agent → it can deflate the compaction threshold.**
With no `curator_model` configured (the default), the extractor runs on `agent._llm` (`fast_extractor.py:148-157`) and appends a turn to the *same* `usage_accumulator`. Compaction reads `turns[-1].current_context_tokens` for its threshold (`context_compaction.py:283-288`). A fire-and-forget extractor turn over a short snippet becomes `turns[-1]` → reports a small context → **compaction can be skipped**. Agent tokens are also double-counted. This is a real SSoT violation on `turns[-1]` that corrupts a *live* feature (compaction), not just memory.

**3. Temporal read-ranking — the core of ADD-only — does not run on the happy path.**
`_freshness_weight`/`apply_policy` run only in the escalation branch (`orchestrator.py:106`); the fast round returns raw RRF order (`:124`). The default `balanced` mode is `max_agentic_rounds=1`, so recency only fires when fast results are `is_weak`. For "where do I work **now**" with good vector hits, recency never runs → a stale fact (Techcombank) ranks equal to the current one (FPT). Also: `valid_until` is populated by FTS (`sqlite_fts_provider.py:59`) but nothing filters expired facts, and `_status_of` is a stub that always returns `"active"` (`fusion.py:85-88`). The PR's "ADD-only + resolved at read by recency" promise isn't realized on the main path.

---

### 🟠 Should fix

**4. Slow lane reads the wrong segment + has no cursor.** It piggybacks at the right trigger (`context_compaction.py:1110-1118`) but passes the full history → takes the **last 40 messages** (`fast_extractor.py:232-245`) = the segment compaction *keeps*, not the middle zone about to be summarized away — the opposite of "extract before detail is summarized." No disjoint-window cursor (spec §4 requires reusing the compacted-segment boundary), so consecutive compactions re-extract overlapping ranges and rely on exact-dedup as a crutch.

**5. Fast lane has no cursor either.** Spec wants a per-session `(cursor, now]` slice for linear cost. The implementation re-sends a fixed last-8-message window with an in-memory `_jarvis_extract_turns` counter (`retrieval_hook.py:148-154`) that resets on restart and isn't keyed per session — inconsistent with the deliberately restart-safe recall change-gate right next to it.

**6. `debug`-level swallows silently lose memories.** `fast_extractor.py` swallows extract/persist failures at `logger.debug` (`:106-108, :132-133, :189-190, :226-228, :269-273`) and the hook at `:199-200`. An entire batch of extracted memories can vanish with nothing above debug level — violates the "no silent fallbacks" rule.

**7. Exact-dedup diverges from spec + three disagreeing dedup keys.** Spec keys on `normalized_content + subject_scope`; the code adds `memory_type` (`memory_service.py:89-100`), so the same fact extracted once as `preference`→semantic and once as `instruction`→pinned won't dedup. There are three different keys — candidate (`candidate_service.py:28`, omits subject_scope), memory (`:43`), conflict `find_similar` — with no authoritative one.

**8. `degraded_reason` is hardcoded `"qdrant_unavailable"`** (`orchestrator.py:145`) even when the backend is Ladybug and Ladybug is what's down → misleading telemetry/banner.

**9. Boot-migration gate `store.count()==0`** (`server.py:474-491`): (a) only enqueues `MemoryRecord`, so `episodic_documents` are never back-filled; (b) if a prior boot populated partially then crashed, `count()>0` means the rebuild never re-runs → permanently partial index. Compare Ladybug-count vs SQLite-active-count, or rely purely on outbox idempotency.

**10. Token attribution uses the wrong column.** There's no `category` column on `TokenUsageRecord`; the `memory:fast`/`memory:slow` sub-tag is smuggled into `agent_name` (`fast_extractor.py:188`). It filters, but it's an SSoT smell vs spec §5. The compactor's own LLM calls (`_llm_one_shot`) aren't tagged at all.

---

### 🟡 Cleanup / nits

- **Dead code the spec said to delete is still present:** `intent_gate.py` (whole module, 0 callers), `intent_router.py` `gate_targets` branch (always `None`), `curator.py` (whole module, only test + self-imports), `conflict.py` `resolve_or_create/find_similar/build_curator/_defer_conflict` (0 prod callers), the `candidate_type=="conflict"` branch (`candidate_service.py:171-186`), and `RECALL/CAPTURE_PROTOTYPES` in `memory_triggers.py`. Spec §6/§9.6 calls for deletion.
- **`RELATES` is dead schema.** The `(Entity)-[:RELATES]->(Entity)` table is created (`ladybug_store.py:88-89`) but never written, and there's no `relations` field in the extractor schema (`fast_extractor.py:33,204`). GraphRAG is MENTIONS-only today. (The PR body is honest — it only claims MENTIONS — but spec §4 wants relations; either remove the dead table or implement it.)
- **Stale/misleading docstrings:** `fts_index.py:1-3`, `qdrant_indexer.py`, `core/database.py:964` still say "production path is Qdrant"; `conflict.py:1-13`/`curator.py` describe write-time curator as if live; `memory_index_worker.py:1` never mentions Ladybug; the worker's "gated by enabled" docstring is inaccurate (it idle-polls every 2s when disabled).
- Entities are dropped on the conflict branch (`candidate_service.py:183` lacks `entities=` vs `:197`) — though that branch is dead.
- Fire-and-forget `asyncio.create_task` with no retained reference (`retrieval_hook.py:152`) — can be GC'd mid-flight.
- **Frontend:** `memory.js` `pendingByAgent`/`pendingCount` is dead state (0 readers — the badge actually runs through the Approvals store); the `memory_indexed` SSE event has a consumer but no producer; two un-i18n'd user-facing strings in `SettingsMemory.vue:230-237` (`stored · hidden`, `not set`).
- **Scope creep:** `useLang.js` flips the default language for new users `vi → en` — a cross-cutting product change bundled into a memory PR; worth a separate sign-off.

---

### Recommendation: **Request changes**

High-quality work with honest test coverage. But three correctness items must be addressed before flipping `memory.enabled` or merging to `main`: (1) the silent writer/reader store divergence, (2) the extractor deflating the compaction threshold via the shared accumulator, and (3) temporal recency — the heart of ADD-only — not running on the happy path. These make actual behavior diverge from the PR's "resolved at read by recency" / "no silent fallback" claims.

Since the flag defaults off, this isn't a block-at-all-costs, but I'd gate merge on the 🔴 group plus #4 and #6. The dead-code / stale-docstring cleanup can be a follow-up, but soon — it currently misrepresents which features actually exist.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

omnigentx:
Thanks — excellent, accurate review. Addressed all **gate-on** items in `e4e8a2b`. Memory/retrieval/compaction suite: **217 pass**.

### 🔴 Must-fix — done
- **#1 writer≠reader SSoT** — worker `_qd()` now DEFERS the dense index (returns `LadybugIndexer(store=None)` → `is_available()==False`) when Ladybug can't open, **symmetric** with the reader's FTS degrade. No more silent divert-to-Qdrant the reader can't see.
- **#2 shared accumulator** — `build_extractor_generate_fn` now ALWAYS builds a separate LLM instance (resolves the agent's own model via `default_request_params.model` on inherit) → own `usage_accumulator`, can't become the agent's `turns[-1]` / deflate compaction / double-count.
- **#3 recency on happy path** — `fusion.apply_policy(now=...)` now runs right after `_fast_round` (not only on escalation) + a **newer-wins score tiebreak** for same-session supersede. New regression test: newer fact outranks older on the balanced path.

### Gate-on extras — done
- **#4** slow-lane now extracts `history[:-keep_recent]` (the segment about to be summarized), not the kept tail; the threshold memo gives disjoint windows.
- **#6** memory-losing extract/persist failures → `logger.warning` (kept `debug` only for telemetry + expected gating skips).
- **#8** `degraded_reason` now reflects the configured backend.

### Deferred to follow-up (you flagged these as OK-to-follow-up)
- **#5** per-session fast-lane cursor, **#7** unify the three dedup keys (drop `memory_type` from the memory key), **#9** migration gate vs partial index, **#10** real `category` column + tag the compactor's own LLM calls.
- 🟡 dead-code deletion (`intent_gate`/`curator`/`conflict` write-path/prototypes), stale docstrings, `RELATES` dead table, frontend nits, and the `useLang` default-language flip (agreed it deserves separate sign-off — I can pull it out of this PR if you prefer).

Will open a tracking issue for the deferred set unless you'd rather they land here before merge.

omnigentx:
Round 2 — finished the **deferred** set too (`0a0f355`). Backend memory suite green, frontend 10/10, full collect clean (1976 tests, no broken imports from the deletions).

### Now done
- **#7** exact-dedup keyed on `owner + normalized_content + subject_scope` (dropped `memory_type`) — same fact as semantic vs pinned now collapses.
- **#10** real `category` column on `token_usage` (+ idempotent ALTER): extraction tagged `memory:fast`/`memory:slow`, **and the compactor's own `_llm_one_shot` is now tracked + tagged `compaction`** (was untracked). No more smuggling category into `agent_name`.
- **Dead-code purge** (§9.6): deleted `intent_gate.py`, `curator.py`, `conflict.py` (moved the one live helper `_ctx_with_override` into `fast_extractor`), the candidate `"conflict"` branch, the `RECALL/CAPTURE` prototypes, and the unused `RELATES` table — plus their orphaned tests (−726 LOC).
- **useLang flip pulled out** — English stays the COMPLETE fallback base, but the default UI language stays `vi`; the `vi→en` default flip is no longer in this PR (separate sign-off).
- i18n'd the two `SettingsMemory` strings.

### Still intentionally deferred (low-impact, noted)
- **#5** per-session fast-lane cursor — the debounce counter resets on restart, but it's a cost-bound only (re-extraction is exact-deduped), not correctness. Will key it per session in a follow-up.
- **#9** migration gate — the `count()==0` path is correct for the realistic case (memory ships OFF; no legacy Qdrant-indexed data exists), and a partial index self-heals via the continuously-draining worker (pending outbox events). A `force` re-enqueue for an actual Qdrant→Ladybug switch is a follow-up.
- stale docstrings + `memory.js` dead `pendingByAgent` — cosmetic, follow-up.

Ready for re-review.

omnigentx:
## Review — fast-agent spawn (agent identity uniqueness)

> **Scope note:** these findings are about the **`backend/fast-agent` submodule** spawn changes (agent-name identity), which are not yet committed and therefore not part of this PR's diff. Posting here for tracking since it's bundled with the `feat/memory-v2` work. Verified locally: files parse, `test_agent_name_uniqueness.py` + `test_agent_creation_uniqueness_audit.py` → 9 passed.

**Positives:** consolidating the three "taken names" sources (live registry + DB team sessions + `agent_definitions`) into a single `_collect_taken_names` is a real SSoT win. Good *why* comments. The static audit test (`test_agent_creation_uniqueness_audit.py`) that blocks adding a creation entry-point without a gate is a nice guardrail.

### Should fix

**1. In-batch name collision in `spawn_team` pre-register loop — `team_spawner.py:552-565`**

```python
for role_name, role_config in roles.items():
    agent_name = _generate_unique_agent_name(role_display, registry)
    session.agents[agent_name] = {...}   # in-memory only
session.write_roster()                    # persisted AFTER the loop
```

`_generate_unique_agent_name` → `_collect_taken_names` reads the registry, `_get_store().list_all()`, and the DB — but the name just assigned lives only in `session.agents`. The store was `upsert`ed **before** the loop (when `agents` was empty) and `write_roster()` runs **after**. So two roles sharing the same `role_display` can be handed the **same name**; the second overwrites the first in the dict → **one agent silently lost**.

- Narrow: `role_display` defaults to `role_name.upper()` (dict keys ⇒ unique); only breaks when a template sets the same explicit `role_display` for two roles.
- **Pre-existing** pattern (not introduced by this refactor), but the new "single source of taken names" comment makes it *look* closed when it isn't. The audit test only checks the gate is *referenced*, not in-batch uniqueness.
- Cheap fix: give `_generate_unique_agent_name` an `also_exclude: set[str]` param and thread the names generated so far through the loop.

**2. Docstring overstates scope — `ensure_unique_agent_name` (`team_spawner.py:123`)**

Docstring claims it's *"the ONE shared uniqueness gate for the three entry points (spawn_agent, spawn_and_run_isolated, **spawn_team_members**)"*, but `spawn_team_members` does **not** call it — it uses `_generate_unique_agent_name`. The audit test itself encodes two different gates. Reword to "two entry points use `ensure_*`; team paths use `_generate_*`".

### Notes (non-blocking)

**3. TOCTOU:** `ensure_unique_agent_name` checks then registers with an `await` in between. Two concurrent spawns with the same explicit name both pass the check. The persistent path has the `agent_definitions` UNIQUE-on-INSERT backstop; the isolated explicit-name path relies on the in-process registry — confirm the registry enforces uniqueness, otherwise there's no atomic backstop there. Low real-world risk (user-supplied names, single event loop).

**4. Perf:** `_collect_taken_names` opens a fresh `sqlite3.connect` + full `SELECT name FROM agent_definitions` on **every** call — N times for an N-member team. Fine for small N; consider caching the taken-set once per batch if teams grow.

_Checked: the isolated path passes `db_path=None` but `_collect_taken_names` falls back to `os.environ["SPAWN_REGISTRY_DB"]`, so `agent_definitions` is still consulted — **not** an asymmetry bug as it first appears._

omnigentx (commented):
## Review — memory v2 (capture · retrieval · indexing)

Reviewed across three layers. The **4 BUGS below I verified by reading the actual post-change source** (file:line cited). The RISK/NIT items are surfaced from a deeper read and cite file:line, but I have not exhaustively re-verified each — treat them as leads. Not approving/requesting-changes (own PR); posting as an honest comment. Top merge-blockers: **B1, B4**.

### 🔴 Bugs (verified)

**B1 — An approved SECRET memory can never persist; it fails silently and diverges candidate/card state.**
`candidate_service.py:58` forces `requires_approval=True` for secrets → user approves → `approve_candidate` (`candidate_service.py:94`) → `_persist_from_candidate` (`:178-182`) calls `create_memory` **without `allow_secret=True`** → `memory_service.py:82` raises `MemoryWriteError`. Via the inbox path (`approval_service.py:367`) the raise is swallowed (`logger.warning` only) while the approval card was already flipped resolved, yet the raise fires *before* `cand.status` is set → candidate stays `PENDING` (card says done, candidate says pending = SSoT divergence). Via the route (`routes/memory.py:215`) it 500s. Either way the explicitly-approved secret is silently dropped. Fix: pass `allow_secret=True` on the user-approved path (not auto-approve).

**B2 — Compactor candidate `subtype` is dropped on persist.**
`candidate_service.py:155` stores `"subtype": c.get("subtype")` in the candidate payload, but `_persist_from_candidate` (`:178-182`) never forwards `subtype=` to `create_memory`, which accepts it (`memory_service.py:69`). Collected, persisted to the candidate, discarded at write. Small data loss.

**B3 — Ledger dedup is keyed by `evidence_id`, which differs per-provider for the same `record_id` → cross-turn double injection.**
`evidence_id` is built as `h.record_id` (ladybug_provider.py:56), `f"memory:{doc_id}"` (sqlite_fts_provider.py:49), `str(chunk_id or id)` (qdrant_provider.py:54). The ledger keys on `evidence_id` (`ledger.py:29,39`) while fusion fuses on `record_id`. So across turns the same memory surfaced via a different provider mix gets a different `evidence_id`, the ledger fails to recognize it as already-injected, and the same fact is injected into context twice — defeating the ledger's stated invariant (`ledger.py:3`). Fix: dedup the ledger by `record_id` (already stored at `ledger.py:16`).

**B4 — Episodic prune leaks the dense node (+ test gap).**
`EVENT_EPISODIC_PRUNE` (`memory_index_worker.py:118-121`) only calls `fts_delete`; it never calls `qd.delete_by_record`, whereas `EVENT_MEMORY_DELETE` (`:122-128`) does. Since `_index_episodic` writes a dense node, a pruned episodic doc stays permanently dense-searchable — an asymmetric delete path / read-after-prune SSoT violation. The test `test_retention_prunes...` only asserts the outbox row is enqueued; it never drains it through the worker to assert the dense entry is gone — which is exactly why this slipped through. Fix: add the symmetric `qd.delete_by_record(row.aggregate_id)` block + extend the test to drain `process_pending` and assert dense removal.

### 🟡 Risks (leads, not exhaustively verified)

- **R1** — fast-extractor hardcodes `subject_scope="user"` (`fast_extractor.py:143`) while compactor uses `agent:<owner>` (`candidate_service.py:154`); exact-dedup keys on `subject_scope` (`memory_service.py:92`), so the same fact captured by both lanes won't dedup. Likely intentional but undocumented.
- **R2** — `_fast_round` swallows provider exceptions (`orchestrator.py:124-129`); if FTS throws, the user gets "0 memories" with no `degraded` flag (computed only from dense availability at `:148`) → total failure masquerading as empty result.
- **R3** — `requires_curator` branch is still honored (`candidate_service.py:74`) but has no caller and no curator service on this branch → a candidate created with it would sit `PENDING` forever. `MemoryStatus.PENDING_APPROVAL` (`models.py:26`) is defined with zero readers.
- **R4** — stale comments/docstrings still reference the deleted `intent_gate.py` (`intent_router.py:45`, `helpers/memory_triggers.py:26,59`, `test_memory_router.py:27`); the `gate_targets` param is dead on the real orchestrator path.

### ✅ Strengths (confirmed)
- Writer/reader vector-backend selection is **symmetric** (SSoT holds); `_qd()` explicitly refuses the "divert writes to Qdrant" trap.
- Integration tests run against the **real ladybug engine** (not mock-only); genuine happy-path e2e `capture→index→recall` (`test_memory_pipeline_e2e.py`) with cross-agent isolation asserted end-to-end.
- The Ladybug "no SET on indexed prop → DELETE+CREATE" constraint is pinned by a contract test.
- RRF math is correct (rank from 1, recency tie-break, empty-list handled). English-first: clean.

omnigentx (commented):
## Review — pass 2 (deep, 5 parallel reviewers: routes · frontend · schema · SSoT-sweep · security)

Follow-up to the first review. NEW findings below (B1–B4 from the prior comment not repeated). I **personally verified the 🔴 items by reading post-change source** (file:line cited); the 🟡 items are credible leads from the reviewers that I have not exhaustively re-verified; the ⛔ section lists claims I **checked and dropped as false positives** so nobody chases them.

### 🔴 Verified (read the code)

**N1 [BUG] None of the three exact-dedup mechanisms has a DB UNIQUE constraint → concurrent inserts duplicate active rows.**
`memory_records.__table_args__` (`core/database.py:764-767`) has only two `Index`, no `UniqueConstraint` on `(owner_agent_name, normalized_content, subject_scope)`; candidate `dedupe_key` and episodic `content_hash` are `index=True` only. Dedup is read-then-insert at app level (`memory_service.py:92-101`, `candidate_service.py:46-71`). Two concurrent captures of the same fact (the extractor fires per-turn) both miss the `existing` check and both `commit()` → duplicate active memories that pollute retrieval ranking. Root-cause fix: partial unique index `WHERE status='active'` on the dedup tuple. (Note: `_dedupe_key` at `memory_service.py:43` is computed but never stored/constrained — dead helper that misleads a reader into thinking dedup is hash-backed.)

**N2 [BUG] The `evidence_id` scheme divergence (B3) also breaks `memory_fetch` — Ladybug/communication evidence is silently unfetchable.**
`memory_fetch` (`services/memory/rpc_handlers.py:70`) does `kind, _, rid = str(eid).partition(":")` and only handles `kind in ("episodic","memory")`. But `ladybug_provider.py:56` emits `evidence_id=record_id` (bare, no prefix) and `communication_provider` emits `comm:{id}`. So evidence surfaced via the graph/comm path yields `kind=<whole-id>` / `"comm"` → matched by nothing → the "view source" fetch returns empty. Fetchability is also non-deterministic (depends on which provider wins fusion). Fix: one canonical `{kind}:{id}` scheme emitted by every provider + `memory_fetch` handling all kinds. (This is the user-visible half of B3; the ledger double-injection is the other half.)

**N3 [RISK · security] A detected secret is stored in plaintext in the candidate payload + the approvals inbox card.**
`_create_approval_row` (`candidate_service.py:206-209`) writes raw `content` into the card `title=f"Remember: {content[:80]}"` and `content=content`. So a secret that `has_secret()` flagged is persisted cleartext into `memory_candidates.payload_json` and `approval_requests.content/title`, then rendered on the Approvals page — exactly what `sensitivity.py` exists to prevent. It never lands in `memory_records`, but it lands on a *more* visible surface. Fix: mask the preview in the card/payload; keep the real value only at the moment persistence is authorized.

**N4 [RISK] `_mem_cfg` NameError on a settings-load hiccup → misleading double log.**
`server.py:457` binds `_mem_cfg = get_memory_settings()` inside a `try`; if it raises, the migration block at `:474` reads `_mem_cfg.enabled` while unbound → `NameError`, swallowed at `:494` as `"[MEMORY] ladybug migration check failed"`. Milder than it first looks — the worker-start failure is already logged at `:468` — but the second log misattributes the cause. One-line fix: `_mem_cfg = None` before the try and guard `if _mem_cfg and _mem_cfg.enabled:`.

### 🟡 Leads (reviewer-reported, not exhaustively re-verified)

- **[RISK · security] No PII detection in `sensitivity.py`.** The `SENSITIVE` tier exists (`models.py:41`) but `classify_sensitivity` only ever returns `SECRET`/`NORMAL` — no email/phone/ID, and bare high-entropy secrets without a `key=`/`token:` prefix are missed. A false negative is persisted *and* recalled into prompts. Wire up the `SENSITIVE` tier + bare-entropy detection.
- **[BUG · UX] Per-row approve/reject/archive/delete have no <200ms feedback and swallow errors** (`frontend/src/views/AgentMemoryPanel.vue:76-92`): no spinner/disabled/try-catch; on failure `error.value` is never set so the user sees nothing. The *bulk* path (same file, `bulk()`) does this correctly — mirror it.
- **[RISK] Candidate badge is dead state.** `pendingByAgent`/`pendingCount` (`stores/memory.js`) is written by `setPending` + SSE but **no UI reads it** (`AgentDetail.vue` tab renders only the label). Either the badge was dropped from the PR or this is dead SSE plumbing.
- **[TEST-GAP] No e2e test for the memory-UI click-path**; `isDegraded` is not seeded on mount (`AgentMemoryPanel.vue:151` vs `loadAll`), so the degraded banner is stale on a fresh tab until the next SSE event.
- **[NIT] Unlogged swallows:** token-usage telemetry (`fast_extractor.py:219`), `entities_json` parse → `[]` (`memory_index_worker.py:31`), `_authorized` on malformed JSON → deny (`communication_provider.py:29`). Each silently degrades with zero observability.
- **[NIT] Uncapped inputs:** `MemorySearchBody.limit` (`routes/memory.py:124`) and `BulkCandidateBody.ids` (`:237`) have no upper bound (contrast the GET routes' `le=200`).

### ⛔ Dropped as false positives (checked, not real)

- ~~"FTS assigns `rank` before filtering → skews RRF fusion"~~ — **not real**: `rrf_fuse` (`fusion.py:36`) re-enumerates by position over the already-filtered list; the stored `bm25_rank` is display-only metadata that RRF never scores on. The rank gaps are harmless.
- ~~"outbox↔Ladybug is not atomic"~~ — true but **by design**: SQLite is the source of truth, Ladybug is a rebuildable projection, and `upsert_memory` is idempotent on re-claim. Only the word "atomic" in the comment overstates it. Not a bug.

### ✅ Confirmed strong
Read-path owner-scoping enforced in SQL (FTS + Ladybug) with provider-layer defense-in-depth; owner never LLM-supplied; FTS injection-safe (`\w+` tokenization, parameterized); locale parity perfect (0 keys missing in en/vi); SSE consumed reactively (not polled); the cross-agent zero-leak eval test is real and runs the actual orchestrator.

---

**Merge-blockers now: B1, B4, N1, N2, N3** + the per-row UX swallow. N1 (no UNIQUE) and the secret path (B1 broken + N3 leaky + no-PII) are the two themes I'd prioritize.

omnigentx:
Addressed the deep multi-reviewer pass in `d6516e6` — every **verified bug** fixed with a regression test:

| Item | Fix | Test |
|------|-----|------|
| **B1** approved secret silently dropped | `allow_secret` on the user-approved persist path only (auto-approve stays locked) | `test_approved_secret_persists` |
| **B2** compactor `subtype` discarded | forward `subtype=` into `create_memory` | `test_compactor_subtype_forwarded_on_persist` |
| **B3** ledger double-injection cross-provider | key the ledger on `record_id`, not per-provider `evidence_id` | `test_ledger_dedups_same_record_across_providers` |
| **B4** episodic prune leaked the dense node | symmetric `qd.delete_by_record` (mirrors `EVENT_MEMORY_DELETE`) | `test_worker_episodic_prune_removes_dense_and_fts` (drains the worker) |
| **N1** no UNIQUE dedup backstop | partial unique index (active rows) built in `init_db` after a one-time dedup pass + `IntegrityError` fallback in `create_memory`; removed dead `_dedupe_key` | `test_memory_dedup_index.py` (migration collapse + index enforcement) |
| **N2** `memory_fetch` empty for graph/comm evidence | canonical `{kind}:{record_id}` via `contracts.evidence_kind` across all providers + `comm` branch (re-checks participant auth) | `test_fetch_comm_evidence_authorized` |
| **N3** secret rendered cleartext on Approvals card | mask the card preview; real value stays in the candidate payload until persistence is authorized | covered via B1 path |
| **N4** `_mem_cfg` NameError in migration block | bind `None` first + guard | `init_db` smoke |
| **R2** dense provider that *threw* read as healthy-empty | track `dense_failed` through `_fast_round` → `degraded` + `{backend}_error` reason | **fixes a pre-existing red route test** (`test_memory_search_route_degraded`) |
| **per-row UX** approve/reject/archive/delete swallowed errors, no feedback | mirror `bulk()`: `rowBusy` set, `:disabled`, try/catch + toast | frontend suite |

Also did the cheap cleanups: **R4** (dropped the dead `gate_targets` path + stale `intent_gate.py` references) and capped `MemorySearchBody.limit` (`le=50`) / `BulkCandidateBody.ids` (max 500).

**Verification:** 299 memory/retrieval/indexing/candidate/router/dedup service tests + the memory-route suite green; `init_db` migration smoke clean; frontend 177 pass; ruff clean. The full `tests/test_routes/` run couldn't be completed locally — it deadlocks when multiple pytest processes share the fixed test DB (an infra artifact, not a code issue); the one failure that surfaced (`test_sessions::test_pagination`) passes in isolation, confirming it was cross-run DB pollution.

**Dropped false positives** acknowledged (RRF rank-before-filter, outbox↔Ladybug non-atomic by design). Remaining **non-blocking leads deferred**: PII tier in `sensitivity.py`, the `pendingByAgent` dead-state / `isDegraded`-on-mount frontend nits, R1 (fast vs slow `subject_scope` — intentional, undocumented), and the unlogged-swallow nits — happy to take any of these in a follow-up.

PR #96, opened 2026-06-19
feat(memory): v2 — LLM recall/capture + GraphRAG on LadybugDB (hardened)

omnigentx:
## Memory v2 → main (LLM-driven recall/capture + GraphRAG on LadybugDB)

Lands the **Memory v2** subsystem into `main` (PR #90 had merged it only into a feature branch, `fix/mobile-ui-floating-widget-scroll`), plus substantial hardening since then. Architecture: **SQLite = source of truth**, **LadybugDB = rebuildable projection** (HNSW vectors + property graph: `Memory`/`Entity` nodes, `MENTIONS`/`RELATES` edges), **SQLite FTS5 = degraded lane**. Retrieval = FTS (BM25) + dense (cosine) + GraphRAG co-occurrence → RRF fusion → relevance gate.

### Highlights
- **Recall pipeline**: FTS + dense + linked-memories fusion, bounded rank-boost policy, calibrated **relevance gate** (off-topic turns recall nothing; threshold configurable).
- **GraphRAG**: per-memory triple extraction `(subject)-[predicate]->(object)`, entity co-occurrence (`MENTIONS`) expansion, configurable `graph_max_hops` over the bipartite Memory↔Entity graph.
- **Event-driven index worker**: transactional outbox + post-commit notify (replaced the 2s polling loop); restart-safe via leases; LadybugDB self-heals a corrupt WAL by rebuilding from SQLite.
- **Knowledge-graph view** on the Memory tab (cytoscape, theme-aware) + restore/archive actions.
- **Compact `memory_search`** output for the agent (dropped debug metadata to save tokens).

### Hardening this round (capture quality + correctness)
- **Dedup across lanes**: the same fact from the agent's `remember` tool and the background extractor now collapses to ONE candidate card (`candidate_type` dropped from the dedupe key; extractor also sees pending candidates).
- **Single-source graph extraction**: `extract_and_store` is the authoritative writer of `relations_json` + `entities_json` (per-memory at persist, every lane), fixing an entity-clobber SSoT bug; `MENTIONS`/`RELATES` share one entity namespace so a mentioned entity and the same entity in a triple are one node.
- **KG migration runs after the agent app is live** (was scheduled before `fast.run()` → extractor LLM unavailable → silent no-op).
- **ADD-only preserved** (per mem0): dedup is NOOP-on-duplicate, never UPDATE.

### Included independent fix (separable)
- **`fix(voice)`**: hands-free "nói chen" (barge-in before the reply lands) no longer deletes the previous user message. Reproduced with a regression test, then fixed; logic extracted to `voiceChatBinding.applyVoiceUserMessage`. *Logically unrelated to memory — flagged here so it can be cherry-picked into its own PR if preferred.*

### Tests
- Backend: 111+ memory tests pass (unit + integration against real SQLite + LadybugDB), incl. an e2e GraphRAG co-occurrence + relevance-gate test through the real orchestrator.
- Frontend: 91/91 `node --test` pass, incl. the new voice barge-in regression test.
- **Gap (stated honestly)**: the live voice click-flow (mic → STT → WS → UI) isn't exercisable headless — verified at the logic layer; needs manual verification.


omnigentx:
## Code review — Memory v2 (PR #96)

Reviewed across 4 subsystems (retrieval, indexing, capture/dedup/KG, integration + voice + frontend). The architecture (SQLite = SoT, LadybugDB = rebuildable projection, event-driven outbox) is sound and test coverage is strong on the happy path. Three findings are worth resolving before merge — each is a case where a **headline guarantee is true single-threaded but untested against its own concurrency/edge failure mode**.

### 🔴 High

**H1 — Dedup-across-lanes has no DB-level guard; duplicate candidate cards can still appear under real concurrency.**
`candidate_service.py:54-79` does a plain SELECT-then-INSERT, and `memory_candidates.dedupe_key` is only a **non-unique** index (`core/database.py:834`) — unlike `memory_records` which has the partial UNIQUE `uq_memory_active_dedup`. The `remember` lane (sync RPC) and the fast extractor (`asyncio.create_task`, `retrieval_hook.py:155`) each open their own session, so two interleaved `create_candidate` calls for the same fact both pass the SELECT and both INSERT → two pending cards. On the default `manual` approval policy the record-layer `IntegrityError` backstop never fires (it only catches at persist/auto-approve), so both cards sit in the inbox. `test_dedupe_collapses_across_lanes` runs the two lanes **sequentially on one session**, so the race is never exercised.
*Fix:* add a partial UNIQUE index on `memory_candidates(dedupe_key)` for open statuses + catch `IntegrityError` and re-SELECT the winner (same pattern already in `create_memory`). Add a concurrent test.

**H2 — `claim_batch` is a non-atomic SELECT-then-UPDATE → double-processing if a second worker/clone runs.**
`outbox_service.py:159-173` selects PENDING rows then flips them to `in_progress` in a separate step. The comment says "atomically claim," but SQLite's default isolation doesn't make read-then-write atomic. Two workers — or the second repo clone the project routinely runs — would both claim the same rows. Most handlers are idempotent (deterministic `point_id`, DETACH+CREATE upsert), but it pairs badly with the single-connection Ladybug store.
*Fix:* claim via a single `UPDATE ... WHERE id IN (SELECT ... LIMIT n) RETURNING *`, or make single-worker a hard guarantee (advisory lock) rather than a comment.

**H3 — `decision.bm25_first` is computed but never consumed; the relevance gate drops exact-identifier recall.**
`intent_router.py:61,69` sets `bm25_first` for queries with an exact identifier (email/ID/code), but `orchestrator.py` only reads `decision.level`/`decision.targets` — `bm25_first` is dead. The gate at `orchestrator.py:150` drops **all** FTS hits whenever the dense lane returns `[]`. Exact identifiers are often semantically far from stored memories, so dense legitimately returns nothing → a strong exact BM25 match is discarded for an on-topic identifier query, contradicting the router's "preserves exact-identifier recall" promise.
*Fix:* when `bm25_first` is true, bypass the dense-empty gate and keep FTS hits. Add a test (identifier query, dense `[]`, assert FTS hit survives).

### 🟡 Medium

- **M1 — Off-topic guarantee is silently conditional on dense health.** "Off-topic turns recall nothing" lives entirely in the dense-healthy-and-empty branch (`orchestrator.py:150`). When dense is **down**, the FTS-only fallback injects function-word hits on off-topic turns; `quality_gate.min_top_score=0.015` doesn't catch a lone FTS hit (RRF `1/61≈0.0164 > 0.015`). Either document this or raise `min_top_score` above the single-list rank-1 score.

- **M2 — WAL self-heal wipes the whole graph on a broad substring match.** `ladybug_store.py:100-108` triggers `_wipe_graph_files` on any `RuntimeError` containing `"wal"` **or** `"corrupt"`, logged at `warning`. A transient/lock error mentioning the wal path could wipe a healthy projection. Tighten to Ladybug's exact signature (`"corrupted wal file"`), log at `error`, and rename-aside instead of `rmtree` so a misfire is recoverable.

- **M3 — Migration rebuild is gated on a fragile `count()==0` heuristic** (`server.py:480`). If a self-heal wipe leaves ≥1 stale node, `count()` is non-zero → no rebuild fires → graph stays permanently inconsistent with SQLite. Memory records carry **no index watermark** (unlike episodic docs, which set `indexed_revision`), so "outbox row = done" can outlive a wiped projection with no way to detect it. Drive the rebuild off a per-record watermark or an in-graph schema marker.

- **M4 — `entities_json` capture-seed is clobbered by an empty extraction.** `extract_and_store` (`knowledge_graph.py:130`) unconditionally writes `entities_json` from triples — when `extract_triples` returns `[]` it writes `json.dumps([])`, destroying a seeded entity list. The documented "fallback survives if extraction fails" is then false. Only overwrite when `triples` is non-empty, or merge seed ∪ triple-entities.

- **M5 — Memories whose first extraction returns `[]` are permanently un-graphed.** `schedule_extract_and_store` gates on `not rec.relations_json`, and `backfill_relations` filters `relations_json.is_(None)`. A valid empty `[]` (non-null) is treated as "already graphed" with no repair path. Distinguish "never attempted" (NULL) from "attempted, empty," or add a `graph_extracted_at` marker.

### 🟢 Low / Nit
- Cache-hit path (`orchestrator.py:100-101`) skips `apply_policy`, so recency ordering freezes at first-call `now` — a hot read-only query serves increasingly stale recency buckets. Cache the pre-policy list and re-apply policy on read.
- `_CACHE` (`cache.py:41-45`) mutates an `OrderedDict` with no lock — safe single-loop, risky if driven from a background thread/loop. Add a lock or document the invariant.
- Dead realtime path: nothing emits `memory_indexed`, so `memory.js:23-25` + `AgentMemoryPanel` `watch(indexTick)` never auto-refresh after indexing. Wire a broadcast from the worker on drain, or drop the branch.
- Degraded banner can stick: `_emit_completed` (`orchestrator.py:197`) returns early on `not evidence and not degraded`, so a recovered-but-empty turn never clears `degradedByAgent`. Emit `retrieval_completed` on any non-degraded turn.
- `vi.json`: `memory.archived/archive/restore` left in English ("Đã lưu trữ"/"Lưu trữ"/"Khôi phục") — violates the bilingual rule.
- `voiceChatBinding.test.js` guards future regressions of the extracted `applyVoiceUserMessage`, but by construction cannot fail against the old inline code — it doesn't reproduce the original defect, just guards the new function. Fine as a guard; don't over-claim.

### ✅ Confirmed good
Curator API key is write-only (`curator_api_key_set: bool`, no plaintext in `get_settings`). No new AF_UNIX socket (delegates over the existing RPC socket — macOS 104-byte limit N/A). `extract_and_store`/`backfill_relations` are genuinely the only writers of `relations_json`/`entities_json` (SSoT consolidated). ADD-only NOOP-on-duplicate confirmed. RRF fusion math + bounded rank-boost + `graph_max_hops` are sound and well-tested. MemoryGraph.vue is theme-aware and bilingual. Memory routes are owner-scoped. KG-migration-after-`fast.run()` ordering fix is a correct root-cause fix. Voice barge-in fix correctly preserves the prior user message.

---
*Reviewed with Claude Code (4 parallel subsystem reviewers). Net: ship-able after H1–H3; the High findings are concurrency/edge gaps the current tests don't exercise.*

backend/services/memory/candidate_service.py
```diff
@@ -0,0 +1,252 @@
...
+    cand = MemoryCandidate(
+        id=uuid.uuid4().hex, owner_agent_name=owner_agent_name,
+        candidate_type=candidate_type, payload_json=json.dumps(payload, ensure_ascii=False),
+        source_refs_json=json.dumps(sources or [], ensure_ascii=False),
+        status=CandidateStatus.PENDING.value, confidence=confidence,
+        requires_curator=1 if requires_curator else 0,
+        requires_approval=1 if requires_approval else 0,
+        dedupe_key=dedupe, created_at=now,
+    )
+    db.add(cand)
```
omnigentx, line 87: **🔴 High — dedup-across-lanes is unguarded; duplicate candidate cards can still appear under concurrency.** This is SELECT (L56) then INSERT (here) with no DB-level uniqueness — `memory_candidates.dedupe_key` is only a non-unique index (`core/database.py:834`), unlike `memory_records.uq_memory_active_dedup`. The `remember` lane (sync RPC) and the fast extractor (`asyncio.create_task`) open separate sessions, so two interleaved calls for the same fact both pass L56 and both INSERT → two pending cards. On the default `manual` policy the record-layer `IntegrityError` backstop never fires. `test_dedupe_collapses_across_lanes` runs the lanes sequentially on one session, so the race is untested. Fix: add a partial UNIQUE index on `dedupe_key` for open statuses + catch `IntegrityError` and re-SELECT the winner (same pattern as `create_memory`).

backend/services/indexing/outbox_service.py
```diff
@@ -0,0 +1,266 @@
...
+    now: float,
+    lease_seconds: float = DEFAULT_LEASE_SECONDS,
+) -> list[MemoryIndexOutbox]:
+    """Atomically claim up to ``limit`` due pending rows: flip them to
+    in_progress with a lease so a crash mid-batch is recoverable. Ordered by
+    ``next_attempt_at`` so retries don't starve fresh work indefinitely."""
+    rows = (
+        db.query(MemoryIndexOutbox)
+        .filter(
+            MemoryIndexOutbox.status == PENDING,
```
omnigentx: **🔴 High — `claim_batch` is a non-atomic SELECT-then-UPDATE → double-processing if a second worker/clone runs.** This SELECT of PENDING rows and the subsequent flip to `in_progress` are not atomic under SQLite default isolation, despite the "atomically claim" wording. Two workers — or the second repo clone routinely run on port 8000/8001 — would both claim the same rows. Handlers are mostly idempotent, but it pairs badly with the single-connection Ladybug store. Fix: claim via one `UPDATE ... WHERE id IN (SELECT ... LIMIT n) RETURNING *`, or enforce single-worker with an advisory lock rather than a comment.

backend/services/retrieval/intent_router.py
```diff
@@ -0,0 +1,87 @@
...
+        return RouteDecision(LEVEL_FAST, "agent requested retrieval",
+                             targets=targets - {TARGET_EXTERNAL}, bm25_first=ident)
+
+    # Fresh external info is NOT memory — a lone external signal stays Level 0.
+    real_targets = targets - {TARGET_EXTERNAL}
+    if not real_targets and not ident:
+        return RouteDecision(LEVEL_NONE, "no historical-knowledge signal")
+
+    reason = "identifier match" if ident and not real_targets else signal
+    return RouteDecision(LEVEL_FAST, reason, targets=real_targets, bm25_first=ident)
```
omnigentx, line 69: **🔴 High — `bm25_first` is set here but never consumed by the orchestrator.** It's computed for exact-identifier queries (email/ID/code) but `orchestrator.py` only reads `level`/`targets`. Combined with the dense-empty gate at `orchestrator.py:150`, an on-topic identifier query whose dense lane returns `[]` (common — identifiers are semantically far from stored memories) gets its strong exact BM25 match discarded, contradicting the "preserves exact-identifier recall" promise. Fix: when `bm25_first` is true, bypass the dense-empty gate and keep FTS hits; add a test.

backend/services/retrieval/orchestrator.py
```diff
@@ -0,0 +1,269 @@
...
+                if dense_on and i == 1:
+                    dense_failed = True
+        # Relevance gate (agentic "retrieve-or-not"): the dense lane is the best
+        # semantic judge — when it's healthy yet returns NOTHING within the
+        # similarity threshold, the query is off-topic, so the FTS lane's hits are
+        # just incidental keyword matches (function words like "gì"/"là" hitting a
+        # memory). Drop them → no memory is injected for an unrelated turn. Only
+        # when dense is genuinely DOWN do we fall back to FTS-only (degraded), which
+        # also preserves exact-identifier recall.
+        if dense_on and not dense_failed and isinstance(results[1], list) and not results[1]:
```
omnigentx: **🟡 Medium — the off-topic relevance guarantee is silently conditional on dense health.** This drop-FTS-when-dense-empty branch is the *entire* mechanism for "off-topic turns recall nothing." When dense is **down** (`dense_failed`/unavailable) the FTS-only fallback injects function-word hits on off-topic turns, and `quality_gate.min_top_score=0.015` doesn't catch a lone FTS hit (RRF `1/61≈0.0164 > 0.015`). Also note `request.types` filtering inside the provider makes a type-filtered-empty dense result look identical to "off-topic." Either document the dense-health dependency here or raise `min_top_score` above the single-list rank-1 score.

backend/services/indexing/ladybug_store.py
```diff
@@ -0,0 +1,432 @@
...
+        dense recall stays dead until a human deletes the files. Since the graph
+        is rebuildable, wipe it and recreate empty on that specific failure; the
+        startup migration (count()==0 → consistency_service.rebuild) re-projects
+        every record from SQLite. Any non-corruption error is re-raised."""
+        import ladybug
+        try:
+            db = ladybug.Database(path)
+        except RuntimeError as exc:
+            msg = str(exc).lower()
+            if "wal" not in msg and "corrupt" not in msg:
```
omnigentx: **🟡 Medium — WAL self-heal wipes the whole graph on a broad substring match.** Any `RuntimeError` whose message contains `"wal"` OR `"corrupt"` triggers `_wipe_graph_files` (L107) of the entire projection, logged at `warning`. A transient/lock error mentioning the wal path, or a future unrelated message containing "corrupt," would wipe a healthy graph — and recovery depends on the `count()==0` rebuild trigger (`server.py:480`), which won't fire if ≥1 stale node remains. Tighten to Ladybug's exact signature (`"corrupted wal file"`), log at `error`, and rename-aside instead of `rmtree` so a misfire is forensically recoverable.

backend/services/memory/knowledge_graph.py
```diff
@@ -0,0 +1,202 @@
...
+    if generate_fn is None:
+        return False
+    db = get_db_session()
+    try:
+        rec = db.get(MemoryRecord, record_id)
+        if rec is None or rec.status != "active":
+            return False
+        triples = await extract_triples(rec.content, generate_fn=generate_fn)
+        rec.relations_json = json.dumps(triples, ensure_ascii=False)
+        rec.entities_json = json.dumps(_entities_from_triples(triples), ensure_ascii=False)
```
omnigentx: **🟡 Medium — empty extraction clobbers a seeded `entities_json`.** This unconditionally overwrites `entities_json` from triples; when `extract_triples` returns `[]`, it writes `json.dumps([])`, destroying any entity list seeded at capture time — so the documented "fallback survives if extraction fails" is false. Separately, a valid `[]` makes `not rec.relations_json` False, so `schedule_extract_and_store` and `backfill_relations` (filters `relations_json.is_(None)`) treat the memory as already-graphed with no repair path. Fix: only overwrite when `triples` is non-empty (or merge seed ∪ triple-entities), and distinguish never-attempted (NULL) from attempted-empty.

omnigentx (commented):
Inline notes for the High/Medium findings (full summary in the issue comment above).

backend/services/memory/candidate_service.py
omnigentx: ✅ Resolved in fa7927d — partial UNIQUE index `uq_candidate_open_dedup` (open statuses) + migration collapsing existing open dups, and `create_candidate` now catches `IntegrityError` → re-SELECTs the winner (mirrors `create_memory`). Concurrent failure-mode test added (`test_dedupe_race_falls_back_to_winner_under_unique_index`). Verified passing locally.

backend/services/indexing/outbox_service.py
omnigentx: ✅ Resolved in fa7927d — `claim_batch` is now a single `UPDATE ... WHERE id IN (SELECT ... LIMIT n) RETURNING id`, so the read-and-flip is atomic and a second worker/clone claims none of the same rows. Test added (`test_claim_batch_does_not_double_claim`). Verified passing.

backend/services/retrieval/intent_router.py
omnigentx: ✅ Resolved in fa7927d — the orchestrator now consumes `decision.bm25_first` and carves it out of the dense-empty gate, so an exact-identifier query whose dense lane returns [] keeps its FTS hit. Test added (`test_bm25_first_keeps_fts_hit_when_dense_empty`). Verified passing.

backend/services/retrieval/orchestrator.py
omnigentx: ✅ Resolved in fa7927d — the dense-down dependency is now documented at the gate, and the FTS lane drops VI+EN function words (`_safe_match_expr` stopword filter) so an off-topic function-word-only query yields nothing even when dense is down and the gate can't run. Test added (`test_safe_match_expr_drops_stopwords`). Verified passing.

backend/services/indexing/ladybug_store.py
omnigentx: ✅ Resolved in fa7927d — self-heal now matches ONLY `corrupted wal file`, logs at `error`, and quarantines (rename to `<path>*.corrupt`) instead of `rmtree` so a misfire is recoverable. Non-corruption errors re-raise (tested). Related: the startup rebuild now fires on `count() < SQLite-active` (not `==0`), so a self-heal that left stale nodes still repopulates.

backend/services/memory/knowledge_graph.py
omnigentx: ✅ Resolved in fa7927d — `entities_json` is only overwritten when `triples` is non-empty, so an empty extraction no longer clobbers a capture-time seed. `relations_json='[]'` is now an explicit terminal 'attempted, empty' state vs NULL 'never attempted' (a failed LLM returns early without committing → NULL preserved → retried). KG tests passing.

PR #97, opened 2026-06-19
feat(i18n): bilingual EN/VI across the frontend (+ flow/zoom + light-theme contrast + key recovery CLI)

omnigentx:
## What

Full **English/Vietnamese i18n** for every current frontend feature, built on the existing `useLang` composable — plus a few UI fixes that surfaced during the sweep and a small backend recovery CLI.

### i18n foundation (`useLang` + locales)
- `t(key, params)` helper: **English is the complete base locale**, other locales are partial overlays (missing key → English fallback → raw key). `{n}`/`{name}` interpolation, reactive on the topbar toggle.
- Locales **auto-discovered** via `import.meta.glob('../locales/*.json')` → a contributor adds a language by dropping one `src/locales/<code>.json` and translating values. **Zero JS changes.**
- `en.json` (base) + `vi.json` (overlay), ~1600 keys.
- **Coverage-guard unit test** (`i18nCoverage.test.js`): fails if any `t('…')` key used in source is missing from `en.json`; reports per-locale translation % (does not fail on overlay gaps).

### Translation
- ~70 components migrated from hardcoded strings (and legacy `lang === 'vi' ? …` ternaries) to `t()` keys.
- Vietnamese translated **contextually, not word-by-word** (two review passes). Product/dev jargon kept in English by glossary: Agent, MCP, Token, Model, Flow, Tree, Conductor, backend, process manager, graceful shutdown, runtime, cron, etc. UPPERCASE status badges + section eyebrows kept English to match the sidebar.
- Fixed a misleading source string: the Setup/Settings "master API key" copy no longer claims it "doubles as the encryption master" (it's the **auth** key; encryption uses a separate `JARVIS_MASTER_KEY`).

### UI fixes (found while reviewing the translated screens)
- **AgentFlow**: rotated the Sankey to a **vertical top→bottom** layout so many agents spread horizontally instead of overlapping in one tall column; added **zoom + pan** (matches OrbitHero); hides the empty Team lanes/labels when there are no teams.
- **Light-theme contrast**: `SettingsVoice` (TURN steps) and `TeamMonitor` used dark-only orphan tokens (`--text-secondary/-body/-sub`) with no light override → faint text (~1.6:1). Switched to themed tokens (`--text-dim`/`--text-muted`), now ~9.6:1.

### Backend (separate commit)
- `scripts/show_api_key.py`: prints the effective `JARVIS_API_KEY` (env/.env → `system_config` DB, same resolution as boot) so a forgotten auth key is recoverable without hand-querying SQLite. Bare key on stdout (pipes to `pbcopy`); diagnostics on stderr.
- Clarified `rotate_master_key.py` header: it rotates the *encryption* master and is **not** a recovery tool.

## Testing
- `npm run test:unit` — **169 pass / 0 fail** (includes the i18n coverage guard).
- `vite build` — green.
- `show_api_key.py` verified against the real DB (output matches `sqlite3`) on both env-set and DB-fallback paths.
- **Gap (honest):** no new Playwright e2e for the language toggle / flow zoom — verified via unit guard + build + manual refresh. Worth adding a toggle→assert-VI spec follow-up.

## Notes for reviewers
- Coordinates with the unmerged `feat/memory-v2` branch (same `t()` system, disjoint namespaces) → low-conflict merge.
- Section-header / badge / a few jargon decisions are deliberate (English) — easy to flip per term by editing `vi.json`.


omnigentx:
## Code Review — PR #97 (i18n EN/VI + flow zoom + key-recovery CLI)

Reviewed the substantive logic (`useLang.js`, `i18nCoverage.test.js`, `AgentFlow.vue`, `show_api_key.py`) and spot-checked the locale files. The bulk of the +5.7k diff is mechanical string migration and reads cleanly. Overall this is a solid, well-documented change.

### What it does
- Turns `useLang` into a real i18n layer: `t(key, params)` with English as the complete base locale and other locales as partial overlays (active → English → raw key). Locales auto-discovered via `import.meta.glob`, so adding a language is one JSON file, zero JS.
- A `node --test` coverage guard fails if any `t('…')` key is missing from `en.json` and reports per-locale translation %.
- ~70 components migrated off hardcoded strings / `lang === 'vi' ? …` ternaries.
- `AgentFlow.vue` rotated to a vertical Sankey + zoom/pan; empty team lanes hidden.
- `scripts/show_api_key.py` recovers the effective `JARVIS_API_KEY`; `rotate_master_key.py` header clarified.

### Verification I ran
- `node --test i18nCoverage.test.js` → **2/2 pass**, vi reported **91% translated (147/1602 keys fall back to English)**.
- Confirmed `show_api_key.py`'s raw `row.value` read is correct: the auth key is persisted with `is_secret=False` (`routes/setup.py:245,270`), so it is plaintext in `system_config` — no decryption needed, SSoT claim holds.

### Findings

**1. `t` identifier collision in `AgentFlow.vue` (maintainability / latent bug)**
This PR adds `const { t } = useLang()` to a template that already uses `t` as a `v-for` loop variable in four places (`v-for="(t, i) in teamsLaid"`, lines 233/242/294/315). It works today only because every `t('…')` translation call sits *outside* those loops — inside them, `t` resolves to the team object. This is a trap: the next person who adds a `t('…')` label inside a team loop will silently call the team object as a function. Suggest renaming the loop var to `team`/`tl` to remove the shared identifier.

**2. 147 vi keys silently fall back to English**
The overlay design is intentional and the guard surfaces the %, which is good. But the PR title says "bilingual EN/VI across the frontend" while ~9% of keys render in English under the vi locale. Worth a quick pass over the 147 to confirm none are prominent user-facing strings (vs. the intentional English glossary terms). Not a blocker.

**3. No Playwright e2e for the toggle / flow zoom (project-convention gap)**
The PR honestly discloses this. Per the project's "Happy Path FIRST" rule, user-facing changes should ship with ≥1 e2e exercising the real click-flow, and the frontend already has a Playwright harness (`frontend/tests/e2e`, chromium-desktop, serves `dist`). A `toggle → assert VI string` spec and a basic flow zoom/pan smoke test would close the gap. The unit guard covers key existence, not rendering.

### Minor / nits
- `show_api_key.py` reads `system_config` via a direct `SystemConfig` query rather than `config_service.get()`. Fine for a recovery script (avoids the decrypt dependency), but it is a second read path for the same row — acceptable here since `is_secret=False` guarantees they can't diverge; worth a one-line comment noting why the direct read is safe.
- `t()` returns the intermediate object if a non-leaf key is passed (e.g. `t('flowLegend')`); not currently hit, but a defensive `typeof === 'object' → key` guard would be cheap.
- Default UI language stays `vi` — deliberate and documented; just flagging given the English-first repo mandate.

### Risk assessment
Low. Logic changes are confined to `useLang`/`AgentFlow`; the rest is string substitution guarded by the coverage test. No security concerns (the recovery CLI keeps the key on stdout / diagnostics on stderr, reads plaintext config). Main residual risk is the absent e2e and the `t`-shadowing trap.

**Recommendation:** Approve after renaming the `AgentFlow` loop var (#1). Add the toggle/zoom e2e (#3) either here or as the disclosed follow-up.

omnigentx:
Thanks for the thorough review. Addressed in `5ebc347`:

**#1 — `t` shadowing in AgentFlow (blocker)** ✅ Renamed the `teamsLaid` `v-for` loop var `t`→`tl` in all 4 loops, removing the collision with `useLang`'s `t()`. (The script-scope `reduce((s, t) => …)` on line 52 is a separate, harmless param — left as-is.)

**Nit — `t()` non-leaf guard** ✅ `t()` now only treats leaf **string** values as translations; a bare namespace key (e.g. `t('flowLegend')`) falls through to the raw key instead of `[object Object]`.

**Nit — `show_api_key.py` direct read** ✅ Added a comment noting the direct `SystemConfig` read is safe because the auth key is `is_secret=False` (plaintext), so it can't diverge from the writer.

**#2 — 147 vi fallbacks** ✅ Audited. 126 are short UPPERCASE badges/eyebrows; the 21 sentence-like ones are all intentionally English — proper nouns (Google/Gmail/Calendar, the `JARVIS` acronym tagline), code/format strings (cron examples, `server_overrides (JSON)`, `MCP Servers ({n})`), and voice wire-states (`vad on · tts streaming`). No prominent user-facing string falls back unintentionally.

**#3 — Playwright e2e (toggle + flow zoom)** — Not in this push. Agreed it's the right follow-up per "Happy Path FIRST"; will add a `toggle → assert VI` spec + a flow zoom/pan smoke test against the existing `frontend/tests/e2e` harness. Tracking as the disclosed follow-up unless you'd prefer it blocking here.

**Nit — default lang `vi`** — Deliberate (preserves existing behavior); flipping the default for new users is a separate product call, out of scope here.

omnigentx:
## Re-review after the conflict fix — ✅ clean

Re-reviewed `5ebc347 fix(review)` + `396ba55 merge main`. Both my blocking note and the merge resolution check out.

### Review findings — all addressed
- **#1 `t` shadowing** → loop var renamed `t`→`tl` across all four `teamsLaid` v-fors in `AgentFlow.vue`. No `t`-as-object collision remains.
- **leaf-string nit** → `useLang.t()` now does `typeof active === 'string' ? … : typeof base === 'string' ? … : key`, so a bare namespace key returns the raw key instead of `[object Object]`.
- **show_api_key.py** → added the comment explaining the direct `SystemConfig` read is safe (`is_secret=False` → plaintext, can't diverge from the writer).

### The merge with #96 was the real risk — verified clean
#96 (now in main) had *independently* turned `useLang` into an i18n layer with a different design (static `import en/vi` + manual `LOCALES`), so the conflict was substantive, not cosmetic. The resolution is a correct synthesis, not a "take theirs":
- `LOCALES = { en, vi }` from static imports (keeps it importable under `node --test`) **plus** #97's `import.meta.glob` for *extra* locales, wrapped in try/catch (glob throws when `import.meta.glob` is undefined) with `if (!LOCALES[code])` so the static base wins — no double-registration.
- All 7 conflicted files retain **both** branches' edits. Spot-checked `SettingsView` / `ChatMessages` / `AgentDetail` / `ApprovalsView`: each has #97's `t()` calls *and* #96's memory refs. Bonus: #96's memory tab was folded into the i18n system (`settings.shell.tab.memory` is a `t()` key, present in `en.json`).

### Verification run locally on the merged HEAD
- No conflict markers in `src/`.
- `node --test i18nCoverage.test.js` → **2/2 pass**; `en.json` now **1676 keys** (absorbed #96's `memory` namespace), vi **91%** (147 fall back to EN, by design).
- `npm run build` → **green** (only pre-existing chunk-size / dynamic-import warnings).

### Still open (non-blocking, unchanged from first pass)
- No Playwright e2e for the language toggle / flow zoom — still the disclosed follow-up. Given the merge now also routes #96's memory tab through `t()`, a `toggle → assert VI` spec would cover both PRs' surface at once.

**LGTM to merge.**

PR #98, opened 2026-06-19
feat(memory): evidence-grounded confidence + fail-loud observability + Qdrant removal

omnigentx:
## What this PR does

Started as "fail loud on missing embeddings + remove dead Qdrant" and grew (all cohering around memory) into a fuller pass on **how durable memory captures confidence, surfaces it, and stays observable**. Grouped by theme:

### 1. Fail-loud observability (the original incident)
- Missing FlagEmbedding no longer degrades silently. The dense lane reports it at **three layers**: boot check, worker first-defer/recovery log (no more "4 rows deferred forever, graph empty, no log"), and the settings **index-status** UI (`dense.backend` + `dense.embeddings`).
- `create_memory`'s `confidence` is now a **required** arg — a caller that omits it is a `TypeError` at the boundary, not a silent `0.5`.

### 2. Qdrant removed; LadybugDB is the only vector/graph backend
- Dead Qdrant path fully deleted; writer≠reader SSoT trap avoided (worker `_dense()` **defers** rather than diverting writes). Delete/prune now defer symmetrically on outage (review follow-up) — no read-after-delete leak.

### 3. Confidence: real, evidence-grounded, deterministic
- **Root cause fixed:** confidence never reached a memory (`_persist_from_candidate` called `create_memory` without it) → everything was `0.5` and the ranking's `(conf−0.5)` boost was a permanent no-op. Threaded end-to-end through the single chokepoint.
- The extractor no longer returns an **arbitrary confidence number**. It returns a verbatim `evidence_excerpt` + `reasoning_type`; the backend **verifies the quote appears in the snippet** and derives confidence with a pure function (`services/memory/confidence.py`). Fabricated/missing evidence → **never auto-saved**, routed to approval (fail-loud). Chosen over a second LLM verifier (predictable + testable + free > non-deterministic + costly).
- The verified excerpt is stored as durable provenance in `memory_sources` (was empty before); `confidence_method` is recorded in `MemoryVersion.metadata_json` for audit/migration. `agent_remember` derives from authority (0.6) instead of flat 0.5.
- One-time `scripts/backfill_memory_confidence.py` recovers real confidence for pre-fix memories from their candidates (no LLM; idempotent, `--dry-run`).
- Spec divergence (additive rank-boost vs multiplicative weight) documented inline in `fusion.py` — intentional (relevance must always dominate).

### 4. Debug UI — retrieval provenance on the recall chip
- The chat "memories used" chip now shows, per recalled memory, **which lanes** surfaced it (`fts`/`dense`/`graph`) + **raw scores** (`rrf` relevance, `conf`). Carried on a durable side-channel (persisted with the recall block, never reaches the LLM as content, zero prompt tokens). Memory-tab search shows the same.

### 5. UX — explain review under auto-save
- When fail-loud routes an unverifiable capture to approval **despite auto-save being on**, both the agent Memory tab and the central Approvals inbox now show **why** (localized reason code: `unverified_evidence` / `secret`).

### Also
- `memory_forget` id-prefix fix (evidence-id `memory:<id>` vs raw record id).

## Tests
243+ backend tests green: fabricated-evidence-blocks-autosave, missing-evidence→approval, confidence→version-metadata, delete/prune-defers-on-outage, fail-loud-logs-once-then-recovery, forget-prefix + owner-scoping, lane/score channel → display. Frontend store tests green.

## Not in this PR (tracked separately)
- The prod deploy fix (`--extra memory` in `backend/Dockerfile` so FlagEmbedding ships) — base image `1.3.0` built/pushed; merge + redeploy applies it.
- Per-relation confidence on RELATES — deliberately deferred (retrieval uses MENTIONS co-occurrence, not RELATES, today).


omnigentx:
## Code review — fail-loud on missing embeddings + remove dead Qdrant (PR #98)

Reviewed the full diff + grepped the branch for leftover references and verified i18n/test scope. Clean, well-scoped PR. The Qdrant removal is complete and the fail-loud observability is correctly layered.

### ✅ Verified good
- **Complete Qdrant removal**: no live `qdrant_indexer` / `qdrant_provider` / `point_id` / `vector_backend` references remain in prod code (only 2 historical comments). Both modules deleted; `pyproject` drops `qdrant-client` + the docker comment block.
- **Fail-loud is layered and cheap**: provider logs `error` once (names the missing `memory` extra), startup logs `error` once when memory is enabled but embeddings are unavailable, worker `_note_dense_state` logs `warning` on first defer + `info` on recovery. `is_available()` doesn't load the 2.3GB model.
- **`index-status` reports the real backend**: `dense = {backend, reachable, embeddings, points}`; the UI ANDs `reachable && embeddings` and shows a distinct `denseNoEmbeddings` vs `denseDown` warning — exactly the incident signal.
- **Settings removal is migration-safe**: dropping the two keys from `_SCHEMA` means stale DB rows are simply ignored (no crash, no migration). `MemorySettingsPatch` drops `qdrant_url`; Pydantic ignores extras, so a stale client doesn't break.
- **i18n parity**: all 5 `settings.memory.dense*` keys present in en + vi, no leftover `qdrant*`, every `t(...)` ref resolves.

### 🟡 Findings (both Low — not merge-blocking)

**Low — `test_index_status_route` doesn't assert the new `dense` shape** (`test_memory_settings_route.py:78`). It only checks `outbox`/`episodic_documents`. But `dense.embeddings` is the core observable this PR adds to surface the incident — the one thing worth a regression assert. Add `assert body["dense"]["backend"] == "ladybug"` and an `embeddings` key assertion.

**Low — `test_memory_graph_route_unavailable_for_non_ladybug` is now stale** (`:93`). The "non-LadybugDB backend short-circuit" it claims to exercise was *removed* in this PR. It now passes only because `SimpleNamespace(enabled=True)` lacks `ladybug_path` → `AttributeError` → caught → empty payload — not the intended path. The name + comment give false confidence. Rename/rewrite to "graph unavailable when the store can't be opened," or drop it.

**Nit — `index-status` probe calls `get_ladybug_store`** (a read-only health endpoint). Harmless in practice because the store is a process-wide singleton already opened at startup, so the probe reuses it. But if the startup open is ever skipped/failed, the probe becomes the first opener and could trigger the quarantine self-heal as a side effect of viewing settings. Worth a one-line note that the probe must not be the first opener.

### Scope note (not in this PR)
The actual prod fix (the empty graph) is adding `--extra memory` to `backend/Dockerfile` so `FlagEmbedding` ships in the image — correctly called out as follow-up. This PR only makes the UI show the right reason; track that deploy change separately.

---
*Net: ship-able. Tighten the two Low test findings so coverage matches the behavior this PR is about.*

backend/tests/test_routes/test_memory_settings_route.py
```diff
@@ -97,7 +97,7 @@ def test_memory_graph_route_unavailable_for_non_ladybug(client, monkeypatch):
 
     import routes.memory as rm
     monkeypatch.setattr(rm, "get_memory_settings",
-                        lambda: types.SimpleNamespace(enabled=True, vector_backend="qdrant"))
+                        lambda: types.SimpleNamespace(enabled=True))
```
omnigentx: **🟡 Low — this test is now stale.** The "non-LadybugDB backend short-circuit" it claims to exercise was removed in this PR (the route no longer branches on `vector_backend`). It now passes only because `SimpleNamespace(enabled=True)` lacks `ladybug_path` → `AttributeError` inside the `try` → caught → empty payload — not the intended code path. The name + comment give false confidence. Rename/rewrite to assert "graph unavailable when the store can't be opened" (e.g. monkeypatch `get_ladybug_store` to raise), or drop it. Separately, `test_index_status_route` (L78) only asserts `outbox`/`episodic_documents` — it should assert the new `dense` shape (`backend == "ladybug"` + an `embeddings` key), since `dense.embeddings` is the core observable this PR adds.

omnigentx (commented):
Two Low test-quality notes inline; full review in the issue comment above.

omnigentx:
## Code Review — fail loud on missing embeddings + remove dead Qdrant

Reviewed the full diff (+1220/−550, 60 files). CI green, mergeable clean. Verified Qdrant is fully gone from production code and `create_memory`'s now-required `confidence` has exactly one production caller (`_persist_from_candidate`), which passes it — so no break.

### Strengths
- **Fixes the root cause + makes it observable** at three layers (boot, worker first-defer/recovery, settings UI). Exactly the "fail loud" the incident needed.
- **Qdrant removal is clean and complete**; the writer≠reader SSoT trap is explicitly avoided (worker `_dense()` defers rather than diverting writes).
- **Confidence as a pure function** is the right call over a second LLM verifier — and the spec divergence (additive rank boost vs multiplicative weight) is documented inline in `fusion.py`.
- **Lane channels are durable** (survive reload, never reach the LLM as content) and tested end-to-end.
- Strong tests: fabricated-evidence-blocks-autosave, fail-loud-logs-once-then-recovery, forget-prefix + owner-scoping, confidence→version-metadata.

### Issues & risks

**1. (Medium) Delete/prune skips dense removal when the dense leg is down → read-after-delete leak.**
`memory_index_worker.py` `EVENT_MEMORY_DELETE` (L159-165) and `EVENT_EPISODIC_PRUNE` (L149-158): if `qd.is_available()` is False, the dense delete is skipped and the row is `return False` → `mark_done`. Unlike the **upsert** path (which now correctly *defers* via `_note_dense_state`), a delete landing during a dense outage is **permanently lost** — the SQLite record is gone but its dense node survives until a full rebuild, so an archived/pruned memory stays semantically searchable. Pre-existing (diff only renames `_qd()`→`_dense()`), but it directly contradicts this PR's read-after-prune SSoT theme (commented at L152-154) and now stands out next to the symmetric upsert defer. Suggested follow-up: `return True` (defer) when `not qd.is_available()` on DELETE/PRUNE so the removal retries on recovery.

**2. (Low) `index_status` opens the LadybugDB store on every settings-page load** (`routes/memory.py` L176-184). The comment correctly notes the probe "must NOT be the first opener" (quarantine self-heal side effect), but if startup's open failed (the very corruption case), viewing settings becomes the first opener and can trigger self-heal as a side effect of a read-only check. Edge case; invariant is assumed, not enforced.

**3. (Low) Backfill matches memory→candidate by `(owner, content)`** (`scripts/backfill_memory_confidence.py` L312). Two distinct memories with identical content+owner both get the `max()` recovered confidence. Idempotent + `--dry-run` + one-time, so acceptable — worth a comment on the collision assumption.

**4. (Nit) `_block_scores`**: `final or rrf or 0.0` falls back when `final` is a legitimate `0.0`. Harmless for a debug chip.

**5. (Process) Scope vs. title.** Body frames this as "two fixes," but it also ships confidence derivation, a debug-UI feature, a backfill script, and the forget-id fix. All cohere around memory, but the title under-sells the blast radius.

### Security
Owner-scoping (defense-in-depth) preserved on both retrieval lanes; secret masking intact; `approval_service._record_to_dict` exposes only the `reason` code, not full `metadata_json` (correctly minimal); `bge-m3` is public, no token baked.

### Verdict
**Approve with one follow-up.** Solid, well-reasoned, well-tested fix. The only substantive item is the delete/prune dense-removal asymmetry (#1) — pre-existing, but now the odd one out and squarely in this PR's theme; worth a follow-up or a small `return True`-on-unavailable change.

PR #100, opened 2026-06-22
fix(memory): recall correctness — dense-after-forget-all, realtime chip, query-anchored GraphRAG

omnigentx:
Two related memory-recall fixes found while debugging why prod recalled nothing (e.g. the user's full name) even though the data was stored.

---

## 1. Dense recall silently dies after "forget all memories"

**Symptom:** auto-recall, the UI memory search, **and** the agent's own `memory_search` tool all returned nothing, while `index-status` reported the dense backend healthy (`reachable=true, embeddings=true, points=8`). Local was fine on the same code.

**Root cause:** **LadybugDB 0.17.1 leaves the HNSW vector index DEAD once the `Memory` table is emptied** (`count→0`). Rows inserted afterward exist (valid embeddings, `count>0`) but `QUERY_VECTOR_INDEX` returns nothing — dense recall dies for every new memory until a full rebuild. Reproduced **12/12** on the real engine. **Partial deletes and updates are safe — only deletion-to-empty triggers it** (so "forget all memories" is the trigger). Reproduced end-to-end on prod live: `archive all 8 → points=0 → restore 8 → search n=0`; a graph rebuild (restart) restored it (`n=4/n=5`).

Compounding (follow-up, not fixed here): the empty dense lane is reported `degraded=false` (count>0 read as healthy), and the off-topic gate then discards the FTS keyword lane too → total silent recall failure.

**Fix:** `delete_memory()` rebuilds the vector index (`DROP_VECTOR_INDEX` + `CREATE_VECTOR_INDEX`) the instant the table empties (`count()==0`). Extracted `_create_vector_index()` as the single source of truth for the index name/params. Confirmed 5/5 that the rebuild recovers a dead index. Blast radius (GitNexus): LOW — `delete_memory` has one runtime caller (`delete_by_record`), 0 affected downstream symbols.

**Test:** `test_delete_to_empty_rebuilds_vector_index` — insert 4 → delete all → insert 3 → dense search still returns 3 (returns <3, usually 0, without the fix). Real LadybugDB, no mocks.

## 2. "Memories used" chip only showed after a page refresh

**Symptom:** the recall chip (🧠 N memories used + FTS/DENSE/GRAPH badges + scores) didn't appear during a turn — only after reloading the page.

**Root cause:** the chip is built solely from persisted history fetched on mount (`chat.js fetchHistory` → `recallLanes`/`recallScores`); the streaming turn never inserts the recall block. The realtime SSE `retrieval_completed` was only used to toggle the degraded banner, and didn't carry lanes/scores.

**Fix:** the retrieval hook broadcasts a `memory_recalled` activity event the moment it injects a block, carrying the SAME content + per-line lanes + scores the block persists (derived from the deduped `fresh` set via a shared `_score_dict`), so the live chip equals the reloaded one. The agents store routes it to the chat store, which inserts the block just before the in-flight assistant placeholder (guarded on streaming conversation + matching agent). On the next history fetch the persisted block transparently replaces the live one.

(Also: `chat.js` relative imports now carry explicit `.js` so the store is importable under `node:test`, matching `agents.js`.)

**Tests:** backend asserts `memory_recalled` is emitted with content/lanes/scores on injection (and silent with no evidence); frontend chat-store test asserts the block is inserted reactively before the reply, and ignored when not streaming or for a different agent.

## Click-flow

User says "forget all memories", states new facts, then asks Jarvis to recall them → recall works (was returning nothing until a restart), and the "memories used" chip appears during the same turn without a refresh.

## Follow-ups (not in this PR)

- Resilience: an empty dense lane should not veto the FTS keyword lane.
- Observability: `index-status` should flag "count>0 but vector_search empty" instead of reporting healthy.
- Consider upgrading LadybugDB past 0.17.1.

---

## Test coverage (incl. e2e)

**#1 dense recall after forget-all:**
- `test_delete_to_empty_rebuilds_vector_index` — store-level, real LadybugDB: wipe-all → re-add stays dense-searchable.
- `test_forget_all_then_readd_keeps_dense_recall` — **e2e through the real worker**: `create_memory → outbox → MemoryIndexWorker → EVENT_MEMORY_DELETE → _delete_dense → delete_memory (empties → rebuild) → re-add → LadybugProvider recall`. Only the embedder is faked. **Verified it fails without the fix.**

**#2 realtime chip:**
- Backend: hook emits `memory_recalled` with content/lanes/scores on injection (silent with no evidence).
- Frontend: chat-store test asserts the block is inserted reactively before the reply, ignored when not streaming / wrong agent.
- **Playwright e2e** `chat-memory-recall.spec.ts` — boots the real chat app, sends a message, and delivers `memory_recalled` over a real EventSource mid-turn: asserts the chip paints in the live DOM (no reload) and, expanded, shows the recalled line + lane badge. Full browser round-trip (useRealtimeStream → agents.processEvent → chat.addMemoryRecallBlock → ChatMessages). Passing.

Manual: #1 also verified live on prod (archive→restore→search returns n=0 before the fix; restart restores).

---

## 3. GraphRAG recall pulled irrelevant (tangential) memories

**Symptom:** an unrelated query (e.g. baby's age) recalled a memory about the user's AI career, surfaced by the GRAPH lane.

**Root cause (pre-existing since the GraphRAG lane shipped in #96 — verified by checking out the pre-provenance commit and reproducing identical pulls):** the graph lane expanded from the **vector-hit seeds'** entities. In a personal-memory store almost every fact co-mentions the user's own entity ("Nguyễn Văn Phúc", measured **64% df** — a universal hub), so any query dragged in user-co-occurring memories. Not a recent regression: lane-provenance labelling (`53a36e3`) only made the pre-existing graph pull *visible* (it was previously mislabelled `dense`).

**Fix (query-entity anchoring + automatic hub suppression):**
- `LadybugStore.query_anchored_memories`: graph-expand ONLY from entities the query NAMES (normalized whole-token match); skip hubs (entity df ≥ 50% of the owner's memories, no hardcoded names). No query entity → `[]` → dense + FTS only.
- `LadybugProvider.search` uses it instead of blind seed co-occurrence; `graph_max_hops <= 0` disables the lane. **Did not touch dense / FTS / off-topic gate / RRF.**

**Measured before/after on real data:** query "bé làm gì đã 1 tuổi" → GRAPH hits **2 → 0** (AI-career memory gone); "con mèo của tôi tên gì" → still anchors the cat entity correctly.

**Tests:** store-level (named-entity pull / no-entity→empty / hub skipped / owner scope) + rewrote the orchestrator e2e to the query-anchored contract (a vector-far memory is pulled ONLY when the query names its entity, never dragged in otherwise).

---

## 4. Recall relevance: Qwen3-Embedding + cross-encoder reranker

**Symptom:** short Vietnamese question→fact recall returned tangential memories (e.g. the cat in a "how old is my baby" query).

**Root cause:** the bi-encoder dense lane embeds query and memory SEPARATELY → can't judge fine relevance (measured: bge-m3 ranks "cat" ~as high as "7-month son" for the baby query). A score/RRF threshold can't fix it — the relevant full-name answer sits at the same score as the noise.

**Fix (both, quality-first):**
- **A. Embedding → Qwen3-Embedding-0.6B** (sentence-transformers, dim 1024 = no schema change; bge-m3 still selectable via `embedding_model`). `recall_min_similarity` re-tuned 0.44 → 0.34 over 19 on-topic + 12 off-topic queries. Startup migration re-embeds the store IN PLACE on an embedding-revision change (no manual cleanup; old memory records are unchanged & compatible).
- **B. Cross-encoder reranker** (`bge-reranker-v2-m3` via sentence-transformers CrossEncoder — FlagReranker breaks on transformers 5.x). New precision stage: rerank top-K fused candidates by joint (query, memory) relevance, drop below `rerank_min_score`, order by reranker; `apply_policy` is reranker-aware. Gated main-process load + Null fallback (keeps fusion order on failure).

**Verified live (after auto-migration):** "bé mấy tuổi" → only "con trai 7 tháng" (cat/AI noise gone); "con mèo tên gì" → only the cat (0.97); off-topic → nothing. Added latency ~90–260ms/turn warm (Qwen embed ~30-75ms + rerank ~30-216ms); first recall after a restart pays a one-time ~7s reranker load (lazy).

**Tests:** rerank reorder/floor/no-op + apply_policy reranker-override; settings/factory tests updated for new defaults + per-model dep dispatch. 238 memory tests pass.

**Deploy note:** needs `uv sync` (adds `sentence-transformers`, `einops`); switching the embedding model triggers a one-time full re-embed (startup migration).

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -55,7 +55,9 @@ dependencies = [
 # SQLite FTS5 keyword search. The vector + graph store is LadybugDB (an embedded
 # dependency, not a separate container).
 memory = [
-  "FlagEmbedding>=1.4.0",   # BAAI/bge-m3 multilingual embeddings + reranker
+  "FlagEmbedding>=1.4.0",        # BAAI/bge-m3 (optional, still selectable)
+  "sentence-transformers>=3.0",  # Qwen3-Embedding (default) + bge-reranker-v2-m3 cross-encoder
+  "einops",                      # required by some sentence-transformers models
 ]
 
 [tool.uv]
--- a/backend/routes/memory_settings.py
+++ b/backend/routes/memory_settings.py
@@ -27,6 +27,7 @@ class MemorySettingsPatch(BaseModel):
     enabled: bool | None = None
     mode: str | None = None
     auto_capture_preferences: bool | None = None
+    extract_every_n: int | None = None
     approval_policy: str | None = None
     pinned_token_budget: int | None = None
     evidence_token_budget: int | None = None
@@ -37,10 +38,14 @@ class MemorySettingsPatch(BaseModel):
     embedding_model: str | None = None
     embedding_revision: str | None = None
     reranker_enabled: bool | None = None
+    rerank_model: str | None = None
+    rerank_top_k: int | None = None
+    rerank_min_score: float | None = None
     retention_episodic_days: int | None = None
     retention_retrieval_runs_days: int | None = None
     recall_min_similarity: float | None = None
     graph_max_hops: int | None = None
+    hub_max_df: float | None = None
     trigger_lexicon_overrides: dict | None = None
     quality_gate_thresholds: dict | None = None
 
--- a/backend/server.py
+++ b/backend/server.py
@@ -513,37 +513,21 @@ async def lifespan(app: FastAPI):
     except Exception:
         logger.exception("[MEMORY] ladybug migration check failed")
 
-    # Embedding-model migration: if the configured embedding model/revision differs
-    # from what the graph was last projected with, the stored vectors live in a
-    # DIFFERENT vector space (e.g. bge-m3 → Qwen3-Embedding) — query vectors (new
-    # model) vs stored vectors (old) compare as garbage and recall silently dies.
-    # We WIPE the graph (not re-embed in place): the HNSW index is built at
-    # CREATE_VECTOR_INDEX time and is effectively static (Kùzu/LadybugDB lineage),
-    # so re-embedding nodes in place does NOT reliably re-index them. The wipe makes
-    # count() < SQLite, so the rebuild we enqueue re-projects everything into a
-    # FRESH graph + index with the new model's vectors.
+    # Embedding-model migration: on a model/revision change, WIPE the graph + re-embed
+    # (the HNSW index is static — re-embedding in place leaves stale vectors). The
+    # helper guards on the new model's deps being installed BEFORE wiping (never
+    # self-inflict a dead graph). See consistency_service.migrate_on_embedding_change.
     try:
         if _mem_cfg and _mem_cfg.enabled:
-            from services.config_service import config_service
-            cur_rev = (_mem_cfg.embedding_revision or _mem_cfg.embedding_model or "")
-            prev_rev = config_service.get("memory", "indexed_embedding_revision")
-            if cur_rev and prev_rev != cur_rev:
-                import time as _t
+            import time as _t
 
-                from core.database import get_db_session
-                from services.indexing import consistency_service as cs
-                from services.indexing.ladybug_store import reset_ladybug_store
-                reset_ladybug_store(getattr(_mem_cfg, "ladybug_path", "data/memory_graph"))
-                _db = get_db_session()
-                try:
-                    enq = cs.rebuild(_db, now=_t.time())
-                finally:
-                    _db.close()
-                # Mark AFTER enqueue (the outbox intents persist across restarts, so
-                # a crash mid-drain just re-drains; the marker won't wrongly skip).
-                config_service.set("memory", "indexed_embedding_revision", cur_rev)
-                logger.info("[MEMORY] embedding model changed (%r → %r) → wiped graph + "
-                            "re-embedding %d records with the new model", prev_rev, cur_rev, enq)
+            from core.database import get_db_session
+            from services.indexing import consistency_service as cs
+            _db = get_db_session()
+            try:
+                cs.migrate_on_embedding_change(_db, _mem_cfg, now=_t.time())
+            finally:
+                _db.close()
     except Exception:
         logger.exception("[MEMORY] embedding-revision migration check failed")
 
--- a/backend/server.py
+++ b/backend/server.py
@@ -513,6 +513,36 @@ async def lifespan(app: FastAPI):
     except Exception:
         logger.exception("[MEMORY] ladybug migration check failed")
 
+    # Embedding-model migration: if the configured embedding model/revision differs
+    # from what the graph was last projected with, the stored vectors live in a
+    # DIFFERENT vector space (e.g. bge-m3 → Qwen3-Embedding) — query vectors (new
+    # model) vs stored vectors (old) compare as garbage and recall silently dies.
+    # Both are dim 1024, so we re-embed IN PLACE (rebuild → worker re-upserts each
+    # node with the new model's vector); no wipe/schema change. The under-populated
+    # check above misses this (count is unchanged), so it needs its own trigger.
+    try:
+        if _mem_cfg and _mem_cfg.enabled:
+            from services.config_service import config_service
+            cur_rev = (_mem_cfg.embedding_revision or _mem_cfg.embedding_model or "")
+            prev_rev = config_service.get("memory", "indexed_embedding_revision")
+            if cur_rev and prev_rev != cur_rev:
+                import time as _t
+
+                from core.database import get_db_session
+                from services.indexing import consistency_service as cs
+                _db = get_db_session()
+                try:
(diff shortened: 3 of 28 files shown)


omnigentx (commented):
## Code Review — PR #100 (memory recall correctness)

Reviewed all 4 fixes. **Overall: strong PR** — each fix cites a reproduced root cause, blast radius is contained, and test coverage is unusually good (real LadybugDB store tests, an e2e worker-path test verified to fail without the fix, a frontend `node:test`, and a real-browser Playwright spec). Comments explain *why*, not just *what*. The findings below are mostly risk-surfacing, not blockers.

### 🟠 Medium — worth addressing before merge

**1. Embedding-migration wipes the graph without checking the new model's deps are installed (`server.py` lifespan).**
The migration triggers purely on `cur_rev != prev_rev` and immediately `reset_ladybug_store()` (wipes the projection) + enqueues a rebuild. It does **not** verify that `sentence-transformers` is importable first. Failure mode: a deployment that ships this code but hasn't run `uv sync` will **wipe a previously-healthy bge-m3 graph**, then fail to re-embed (Null provider) → dense recall + KG dead until deps land, with no rollback. Since the whole PR is about *avoiding* silent total-recall failure, this self-inflicted variant is worth guarding: probe `get_embedding_provider(cur_model).is_available()` (or `_have(_dep_for(...))`) **before** wiping, and skip + log loud if the new backend isn't present.

**2. The reranker floor is a new hard gate that can zero out the entire recall set.**
`_apply_rerank` returns only `kept` (candidates ≥ `rerank_min_score`); if every candidate scores below the floor, recall returns `[]` even when dense/FTS found relevant rows. With `bge-reranker-v2-m3` (sigmoid 0–1) a genuinely off-topic query can push everything under the floor → empty. That's the *intended* off-topic→nothing behavior, but it adds another total-failure veto point — the same failure **class** this PR is fixing elsewhere (and your own follow-up note: "an empty dense lane should not veto the FTS keyword lane"). `rerank_min_score = 0.005` keeps the present risk low, but please make this an explicit, conscious gate (a one-line comment that the reranker can legitimately empty recall, and/or a floor-of-last-resort that keeps top-1 when the fused set was non-empty).

**3. The `server.py` embedding-revision migration has no automated test.**
It's the most destructive code in the PR (wipes the graph projection) and the only piece running on every prod startup, yet coverage stops at the store/worker layer. A small test over the lifespan branch — "revision unchanged → no wipe", "revision changed → reset + rebuild enqueued + marker set after enqueue" — would lock in the marker-after-enqueue ordering you call out in the comment.

### 🟡 Low — notes / footguns

**4. `_apply_rerank` silently truncates to `rerank_top_k`.** `cand = fused[:top_k]` and only `kept` (⊆ cand) is returned, so `fused[top_k:]` is dropped from recall entirely, not just from reranking. Fine for memory recall (you inject far fewer than 20), but it's a silent cap — a one-word comment ("tail beyond top_k is intentionally dropped") would prevent a future "where did my candidate go" hunt.

**5. `recall_min_similarity` retune (0.44→0.34) is coupled to Qwen but not enforced.** `embedding_model` is independently configurable; selecting `bge-m3` via config without also bumping `recall_min_similarity` back to ~0.44 leaves the gate mistuned for bge-m3's score scale. The comment flags it, but nothing couples them. Consider deriving the default gate from the model, or validating the pair.

**6. Frontend live-vs-persisted block reconciliation.** `addMemoryRecallBlock` inserts a `mem-live-${uuid}` block; the PR relies on the next `fetchHistory` "transparently replacing" it. That holds only if `fetchHistory` rebuilds the conversation from scratch (discarding live blocks). If anything re-fetches/merges history mid-stream, you'd get a duplicate chip. Worth a one-line confirmation that fetchHistory is a full replace, not a merge.

### ✅ Looks good
- Fix #1 (`delete_memory` rebuild on empty): `count()==0` under reentrant `RLock`, `_create_vector_index` as SSoT, and a real-engine regression test verified to fail without the fix. Clean.
- Fix #3 (query-anchored GraphRAG + hub suppression via df≥50%, no hardcoded names): correct root cause (64% df hub), off-switch via `graph_max_hops<=0`, owner-scoped, with measured before/after. The whole-token space-padded match (`f" {norm} " in qn`) is a reasonable normalized-token guard.
- `_score_dict` as the single source for both the persisted channel string and the live SSE payload — exactly the SSoT fix that keeps the live chip == reloaded chip.
- English-first respected; Vietnamese only as test data (allowed per project convention).

### Test-coverage gaps (acknowledge or close)
- The **real** Qwen3-Embedding + cross-encoder model paths are only exercised by fakes in CI (manual/prod verification noted) — acceptable given model weight, but worth stating in the PR so it's a conscious gap.
- The lifespan migration branch (finding #3) is untested.

omnigentx:
@/tmp/pr100_reply.md

PR #101, opened 2026-06-23
feat(memory): Qwen3 reranker + recall tuning + realtime saved chip + rrf/rerank score split

omnigentx:
Stacked on #100. Four related memory changes, developed/measured incrementally on the live Vietnamese store. (The original title was just the threshold tuning; this body reflects everything that actually ships — per review.)

## 1. Reranker swap — bge-reranker-v2-m3 → Qwen3-Reranker-0.6B
Live head-to-head (2026-06-23, real store) showed **bge is the precision bottleneck**: it compressed every Vietnamese score near 0, so the correct "where do I work" answer scored **0.0013** (buried at #2) — no floor could gate, and ranking was wrong (direct answer ranked behind name/entity matches). An earlier "Qwen worse" result was a synthetic-query artifact.

**Qwen3-Reranker-0.6B** (causal LM scored via the **yes/no next-token logit**, not a SequenceClassification cross-encoder → new `Qwen3Reranker` class, dispatched by model name) ranks the direct answer **#1 @ 0.245** and gives a usable spread (answers ≥0.003, off-keyword ≤0.0009). bge still selectable. Auto-selects cuda/mps/cpu (fp16 on accelerator): **warm rerank ~0.44s on mps**, ~1.2–1.4s p50/p95 CPU end-to-end — a real latency cost on the recall hot path, accepted for the quality gain.

## 2. Threshold tuning
- `recall_min_similarity` 0.34→**0.30** — grid-searched (78-query labelled set, Qwen3-Embedding): on-topic recall@5 is flat 100% across [0.05, 0.39]; 0.34 sat at the 0.40 cliff. DENSE-lane property measured **pre-rerank-gate** (independent of the reranker).
- `rerank_min_score` 0.0→**0.001** — viable only because Qwen gives a spread bge couldn't.

**Live acceptance:** "Angular latest version?" 5 leaked memories → **0**; math → 0; "where do I work?" answer #1; archived "where is my child?" → 0. **Residual:** a memory the reranker scores high for the WRONG query (con/Milo adversarial) still leaks — no scalar floor catches it (intent/LLM, deferred).

## 3. Recall-score split (UI)
The "memories used" chip rendered `final` but labelled it `rrf`; after reranking `final` = the reranker score (≈1.0 on a near-verbatim restate), so it showed the cross-encoder score under the wrong name while the real RRF (~0.03) was never shown. Now persists + renders **rrf and rerank separately**, with legacy 3-field back-compat (old blocks render a neutral "score").

## 4. Realtime "memory saved" chip (user-facing feature)
Recall already shows a live chip; **saving was silent** until the user opened the Memory tab. New `memory_saved` SSE (`_emit_saved`, 3 lifecycle points; secrets masked) → `chat.addMemorySavedBlock` → a Vue chip (design-system tokens: success=saved, warning=pending) with **inline undo / approve / reject**, covering BOTH approval cases (auto-approved vs manual/secret pending), with in-place pending→saved/rejected transitions. Session-only notification (not LLM context, not persisted). Bilingual + Playwright e2e.

## Tests
~55 backend memory tests (settings, candidate `memory_saved` incl. secret-mask, reranker dispatch + scoring-math, recall score-split round-trip) + frontend store unit + 2 Playwright e2e (recall + saved chips), all passing.

## Follow-ups (not here)
- con/Milo adversarial & deep ranking → needs a stronger signal (intent gate / LLM judge).
- Persist the saved chip across reload (currently ephemeral; needs candidate↔session linkage).
- `retrieval_runs.rerank_ms`/`total_ms` telemetry records 0 (pre-existing gap).

### DIFF
--- a/backend/agent.py
+++ b/backend/agent.py
@@ -332,6 +332,11 @@ else:
       call memory_remember to persist it (it self-gates / awaits approval).
     - Use memory_search whenever personalization would improve the answer
       (e.g. trip planning → search for the user's transport, family, budget).
+    - PHRASE the query as a SHORT, NATURAL question that names the subject
+      ("what is the user's job?"), NOT a pile of abstract keywords — recall is
+      embedding-based, so a keyword-stuffed query matches nothing. For a broad
+      need, run SEVERAL focused searches (job, skills, certifications) and merge,
+      rather than one long catch-all query.
 
     OUTPUT FORMAT RULES:
     - Use Markdown when appropriate (heading, bullet, table, code block).
--- a/backend/services/memory/candidate_service.py
+++ b/backend/services/memory/candidate_service.py
@@ -103,6 +103,9 @@ def create_candidate(
         return cand                                  # await curator decision
     if requires_approval:
         _create_approval_row(cand)
+        # Live chat chip: surface the pending memory IN CONTEXT (with inline
+        # approve/reject) instead of leaving it silent until the Approvals tab.
+        _emit_saved(cand, status="pending")
         return cand
     # Deterministic, unambiguous → persist now.
     _persist_from_candidate(db, cand, CandidateStatus.AUTO_APPROVED.value,
@@ -138,6 +141,7 @@ def reject_candidate(db: Session, candidate_id: str, *, now: float | None = None
     cand.resolution_json = json.dumps({"reason": reason}) if reason else None
     db.commit()
     _emit("memory_candidate_rejected", cand)
+    _emit_saved(cand, status="rejected")    # live chat chip: pending → dismissed
     if not _from_approval:
         _close_linked_approval(candidate_id, "reject")
     return cand
@@ -252,6 +256,9 @@ def _persist_from_candidate(db, cand, status, *, now, changed_by="system",
     cand.resolved_at = now
     db.commit()
     _emit("memory_candidate_approved", cand)
+    # Live chat chip: this candidate is now an ACTIVE memory (covers BOTH the
+    # auto-approved path and a human approval) → tell the chat the moment it lands.
+    _emit_saved(cand, status="saved", record_id=(rec.id if rec is not None else None))
 
     # SINGLE-SOURCE graph projection: extract this memory's triples (→ RELATES +
     # MENTIONS) off the hot path, for EVERY lane incl. the agent's free-text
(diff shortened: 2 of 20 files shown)


omnigentx:
## Code Review — PR #101

**Reviewer:** Claude Code · scope: full diff vs `fix/memory-dense-index-rebuild-on-empty` (642 / -51, 20 files)

### Overview
The diff bundles **four** distinct changes, not just the titular threshold tuning:
1. **Threshold tuning** — `recall_min_similarity` 0.34→0.30, `rerank_min_score` 0.005→0.001
2. **Reranker swap** — `bge-reranker-v2-m3` → `Qwen3-Reranker-0.6B`, incl. a new `Qwen3Reranker` causal-LM class (yes/no logit)
3. **Recall score split** — persist + render `rrf` and `rerank` separately (was a single conflated `rel`), with legacy back-compat
4. **Live "memory saved" chip** — new `memory_saved` SSE (`_emit_saved`), chat-store block, Vue chip with inline approve/reject/undo, i18n, e2e

The threshold/score work is careful and well-commented. Two things should block merge as-is, plus some follow-ups.

---

### 🔴 Blocking

**1. The PR description contradicts the code it ships.**
Body **Finding 3** + the "What changed" table state the measured conclusion as: *keep bge-reranker, set `rerank_min_score` = **0.0**, Qwen3-Reranker rejected (it scored off-topic-keyword higher, up to 0.98)*. The actual diff does the **opposite**: switches **to** `Qwen3-Reranker-0.6B` and sets `rerank_min_score` = **0.001**. The body predates commit `96d860c` and is now stale — a reviewer reading it is actively misled about what merges. `settings.py` also disagrees with itself across runs (Techcombank 0.245 in one comment vs 0.167 in the reranker docstring). **Rewrite the body + reconcile the numbers before merge.**

**2. Undocumented scope: the entire "memory saved" chip feature.**
The body never mentions change #4 (a real user-facing feature: new SSE event, store action, component, e2e, locale keys). Either split it into its own PR or document it in the body + title. Bundling a frontend feature under "tune recall thresholds" makes the diff hard to review and hard to revert cleanly.

---

### 🟡 Should address

**3. The 0.30 plateau was grid-searched on a different pipeline than ships.**
`recall_min_similarity` comment says the 78-query grid used "real Qwen3-Embedding + **bge** reranker." The PR ships **Qwen3** reranker + floor 0.001. The dense gate is pre-rerank so 0.30 is largely independent — but the *recall@5 = 100% flat* evidence was measured end-to-end on the old reranker+floor. Re-confirm the headline metric holds with the shipping config, or note explicitly that recall@5 is measured pre-rerank-gate.

**4. Empirical claims are unreproducible; reranker math is untested.**
The 78-query labelled set and the live-measurement harness aren't in the repo, so none of the "measured" claims can be re-run. `test_get_reranker_dispatches_by_model` only asserts the **class** dispatch (lazy, no weights) — the actual `Qwen3Reranker.rerank()` scoring (yes/no logit, left-pad, softmax over `[no, yes]`) has **zero** coverage. Consider committing the harness as a marked/slow test, and a tiny tokenization-scaffold unit test that doesn't need weights.

**5. Latency on the recall hot path.**
`Qwen3-Reranker-0.6B` is a 0.6B causal LM doing a forward pass over up to `rerank_top_k=20` docs (~512 tok each), CPU, main process, on every recall. That's plausibly seconds vs the bge cross-encoder. The comments acknowledge "slower but worth it" but there's no measured latency budget. Please attach a p50/p95 number — recall sits in the chat turn.

---

### 🟢 Minor / nits
- `_emit_saved` wraps everything in `except Exception: pass` (no log). Fine as a UI mirror, but a `logger.debug` would help when the chip silently never appears.
- Secret masking SSoT is **good** — both the approval gate (`candidate_service:75`) and `_emit_saved` call the same `has_secret(content)`, so they can't diverge. ✓
- Manual-curator-policy candidates (`requires_curator`, early `return cand`) emit no `memory_saved` chip — looks intentional ("await curator decision"), just confirming it's by design.
- "autoregressive" in the reranker comment is slightly off — it's a single next-token forward, not generation.

### Tests
17 backend memory tests + frontend store/e2e updated and passing per the body. Coverage is solid for the chip and the score-split back-compat (`len(parts) >= 4` vs legacy 3-field). Gaps noted in #4.

**Verdict:** logic is sound and well-documented; **fix the description↔code contradiction (#1) and the scope/title (#2) before merge.** #3–#5 are follow-ups that should at least be acknowledged in the body.

omnigentx:
Thanks — addressed. Summary:

**🔴 #1 (body↔code contradiction):** PR title + body rewritten to reflect what actually ships (4 changes: Qwen reranker swap, threshold tuning, rrf/rerank split, saved chip). The stale "keep bge / floor=0.0" text is gone; the Techcombank number is reconciled to the live **0.245** everywhere.

**🔴 #2 (undocumented scope):** the "memory saved" chip is now its own section (#4) in the body with the SSE/store/component/e2e surface called out. Kept as one PR (incrementally built on this branch) rather than splitting, but it's no longer hidden.

**🟡 #3 (0.30 measured on bge pipeline):** clarified in the `recall_min_similarity` comment — recall@5 is a DENSE-lane property measured **pre-rerank-gate**, so the bge→Qwen swap doesn't move it.

**🟡 #4 (reranker math untested):** added `test_qwen3_reranker_scoring_math` — weight-free, covers left-pad → last-token → softmax over `[no,yes]` → P(yes) per doc in order.

**🟡 #5 (latency):** measured **~0.44s warm rerank on mps**, ~1.2–1.4s p50/p95 CPU end-to-end. Added cuda/mps/cpu auto-selection (fp16 on accelerator) + the number to the docstring/body. (`retrieval_runs.rerank_ms` records 0 — a pre-existing telemetry gap, noted as a follow-up.)

**🟢 nits:** `_emit_saved` now logs at debug instead of bare `pass`; "autoregressive" → "single next-token forward"; secret-mask SSoT and the by-design curator-skip confirmed.

Commits: `bc3fc57`.

PR #80, opened 2026-06-07
fix(ui): mobile UI/UX pass — responsive, FABs, light-theme, approval-comment lock, key recovery

omnigentx:
Mobile UI/UX pass from manual QA, plus a couple of correctness fixes that surfaced along the way. **Not CSS-only** — includes a JS composable, a backend guard, and tests. Includes the light-theme code-block fix originally in #79 (merged in here), so **#79 can be closed**.

## Responsive layout (mobile)
- **Approvals**: detail header stacks (long title was crushed one-word-per-line); content viewer drops its 320px nested-scroll cap; the whole view is now a **single page scroll** (was a mess of nested queue/detail/preview scrollbars).
- **Story reader**: hide the duplicate in-content back; add **prev/next chapter** buttons in a top controls row; put the play button on the same row as the font-size segment.
- **Notification detail**: hide the in-content "← Back" (app-bar already has one); header stacks so the title isn't crushed and actions get a full-width row.
- **Meetings**: participant roster on one horizontally-scrollable row (was 3–4 wrapped rows); hide the diagnostic footer (SSE status + raw `/stream` URL).
- **MCP Servers**: list header hides eyebrow + long help text and gives "New server" its own row; **Refresh + detail Test/Edit/Delete are icon-only** on mobile; BUILTIN badge no longer clips; STOPPED/BUILTIN wrap under long names; ENV VARS rows reflow (KEY on top, value·eye·remove below) and the reveal-eye sits in a fixed slot so rows align; long tool descriptions wrap.
- **Skills**: header hides eyebrow + help text, stacks "New skill".
- **Team Monitor / Agents**: hide the decorative in-page title + tagline (the app-bar already names the screen). Stat-summary headers (scheduler/stories/skills/mcp/tokens) kept — that's live data, not a redundant title.
- `GlobalVoiceIndicator` mic clears the notch + mini-player via the tabbar-aware `calc(...)`.

## Floating action buttons (chat + voice)
- New `useFabVisibility` composable (SSoT singleton): `visible = !manualHidden && !scrollHidden`.
  - **Scroll**: hide on scroll-down and **stay hidden while reading**; reveal on scroll-up (no idle-reveal that re-covered content).
  - **Grip** (mobile): one-tap hide/show, **sticky + persisted**, works on no-scroll pages; gated to routes where a FAB actually renders.
[...]

PR #104, opened 2026-06-24
fix(security): pre-OSS audit remediation — PII scrub, conversation-scoped memory SSE, secret-detector + voice/ice auth

omnigentx:
Remediates all 5 findings from the pre-OSS multi-agent security + wording audit (baseline ~PR #80 → HEAD). **One PR.**

## Findings fixed

| # | Severity | Issue | Fix |
|---|----------|-------|-----|
| 1 | 🔴 blocker (PII) | Maintainer's real name / employer / district / family used as code examples (incl. shipped to the LLM via `TRIPLE_PROMPT`) | Removed the PII tokens, replaced with neutral placeholders, across the source files + tests + docs that contained them |
| 2 | 🟠 high | `memory_recalled` / `memory_saved` SSE fan out per-agent only → a recall/saved chip from conversation A paints into conversation B (another tab, same agent) | Stamp originating conversation via new `current_conversation_id` ContextVar; gate the chat store on a matching id |
| 3 | 🟡 medium | Secret detector missed URI-userinfo creds (`redis://user:pass@host`) and bare high-entropy tokens → auto-saved cleartext + embedded into the unencrypted index | Two conservative patterns (force approval, never drop data) |
| 4 | 🟢 low | `GET /api/voice/ice` served minted Cloudflare TURN creds with **no auth** (12 sibling routes have it) | `Depends(verify_api_key)` + corrected the false docstring |
| 5 | 🟢 low | `addMemorySavedBlock` had no streaming guard (sub-case of #2) | Subsumed by #2's conversation gating |

## Scope note on #1 (PII vs i18n)
This PR removes **PII tokens** only — it is not a full anglicization. `knowledge_graph.py`'s `TRIPLE_PROMPT` deliberately **keeps** the Vietnamese predicates (`thích`, `làm việc tại`, …) because they are functional for Vietnamese triple extraction ("keep the statement's language"); only the real employer/district/family tokens were swapped for neutral ones. The broader English-first comment/example cleanup is tracked separately (internal i18n backlog, kept local).

## #2 design (the cross-cutting one)
- Backend: `current_conversation_id` ContextVar set in the chat route alongside `current_run_id`. It is **copied into `asyncio.create_task`**, so the async fire-and-forget extractor still emits the right id after the turn (covered by a cross-task test that resets the var before the task runs). `_emit_recall_block` / `_emit_saved` stamp it on the payload.
- Frontend: `belongsToConversation(data, conv)` drops an event only when **both** ids are known and differ — fresh conversations (id not resolved yet) and legacy/voice events with no id keep the old active-conversation routing (back-compat).

## Review follow-ups (2nd commit)
- #3 secret rule tightened to `[A-Za-z0-9]{32,}` so hyphenated UUIDs / underscored record-ids no longer false-positive into forced approval (40-hex tokens still caught). +2 negative tests.
- #2 saved-path cross-task propagation now has its own test (not just the synchronous recall path).
- **Untracked internal dev docs** (kept local via `.gitignore`): the `*-spec.md` files, `docs/memory-impl/`, `memory-v2-plan.md`, the i18n backlog, and the one-time `scan_vietnamese.py`. Only `SELF_HOSTING.md` + `VOICE_WEBRTC_TESTING.md` remain public-facing.

## ⚠️ Follow-up NOT in this PR
The PII (and the now-untracked internal specs) still live in **git history** (7 commits, e.g. #96 already on `main`). Editing/untracking files does not remove them. The repo is still **private**, so a history rewrite/squash before the first public push is feasible — **decision pending** with the maintainer.

## Verification
- Backend: memory + voice/session suites green; new tests — sensitivity (12, incl. UUID/record-id negatives), recall + saved conversation-id stamp, voice/ice auth (2).
- Frontend: `node --test` 197/197 (incl. conversation-scoping), `npm run build` clean.
- e2e: `chat-memory-recall`, `chat-memory-saved`, `team-monitor-v2`, `approvals-mobile-responsive` pass.


omnigentx:
## Code Review — pre-OSS audit remediation

**Verdict: Approve with minor notes.** Five well-scoped fixes, real tests, and honest about what's deferred. I validated the two load-bearing parts — ran the secret regexes live, and traced the SSE conversation-routing through the code. Nothing blocking.

### What's solid
- **#2 mechanism is correct.** Traced it end-to-end: `chat.py:441` sets `current_conversation_id` inside `run_chat()`; the extractor is spawned via `asyncio.create_task` at `retrieval_hook.py:236` **during the turn**, so it snapshots `conv-X` and survives the `finally` reset (a `copy_context()` snapshot, not a live ref). `_emit_saved`/`_emit_recall_block` read it correctly. `belongsToConversation` drops only when **both** ids are known and differ → fresh-conversation and legacy/voice (null id) keep back-compat routing. Clean.
- **#3 regexes work and don't FP on prose.** Ran the new suite (10/10) + my own probes: `redis://user:pass@` caught, plain URLs and ordinary sentences not flagged. "Force approval, never drop" is the right conservative stance.
- **#4** is a legit privilege fix (billable TURN creds were anonymous), with a corrected docstring and a no-auth-rejected test.
- **Honesty:** the PR explicitly flags PII still in **git history** (#96 already on `main`) and defers the rewrite — correct call, and that history rewrite is the actual OSS blocker, not this diff.

### Notes / risks (none blocking)

**1. Secret-detector false positives — confirmed live.** The bare-token rule `[A-Za-z0-9_\-]{32,}` (letter+digit) flags UUIDs, 40-char git SHAs, and record-id-like strings as secrets → all route to manual approval:
```
True  uuid (550e8400-e29b-41d4-a716-446655440000)
True  git_sha40 (4f9a1c8e7b2d3a6f5e0c9b8a7d6e5f4c3b2a1d0e)
True  record_id_like (mem_01HXYZ7Qpqr8stuvwxyz0123456789ab)
False plain_url / normal prose
```
Per design this only **forces approval, never drops**, and the inline comment owns it — so it's an accepted tradeoff. But if users paste IDs/hashes into memories with any frequency, they'll see spurious approval prompts. Worth a conscious maintainer "yes, we accept this". Cheap refinement if not: exclude hyphenated-UUID shape and pure-hex `{40}` SHAs.

**2. Test gap on the saved-path's cross-task propagation.** `test_recall_sse_carries_conversation_id` covers the synchronous **recall** path. The **saved** path — the whole reason #2 needs a ContextVar — relies on `create_task` snapshotting the var into the async extractor, and that hop is asserted only by reasoning, not a test. If the extractor is ever refactored to a persistent worker/thread, the saved chip silently mis-routes again with **no failing test**. Recommend one test that spawns the extractor as a task and asserts `_emit_saved` carries the id. (Side note: `fire_slow_extraction_from_history` emits null id → shows in active conversation; fine as back-compat, but undocumented.)

**3. #4 depends on `verify_api_key` accepting the SPA session cookie.** If it only honored a bearer/API-key header, this would break the voice happy-path for cookie-authed SPA users. The 12 sibling voice routes already use this dependency, so almost certainly fine — just the one thing to confirm, since no e2e exercises the authed `/voice/ice` from the browser flow.

**4. PR-body wording nit.** "Replaced with neutral English placeholders across ~10 source files" slightly overstates it — `knowledge_graph.py` deliberately **keeps** Vietnamese predicates (`thích`, `làm việc tại`) because they're functional for VI extraction; only the PII tokens were removed. That's the right scope decision, but the body reads as a full anglicization. `oss-i18n-backlog.md` documents this correctly.

### Minor
- `scan_vietnamese.py` is a clean, dependency-free helper — good to have it in-repo.
- Fixture swap `Gia Lâm → Hà Nội` de-identifies while keeping the vi+en code-switching coverage intact. Good.

**Bottom line:** ship it. Address the saved-path test as a fast-follow, and get an explicit maintainer ack on (a) the ID/hash → forced-approval tradeoff and (b) the git-history rewrite before the first public push.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

omnigentx:
## Follow-up: drop `scripts/scan_vietnamese.py` from this PR

One more thing I should've caught above. `scan_vietnamese.py` is **one-time audit scaffolding** that landed in `scripts/` — a path that implies a maintained, long-lived artifact. For a repo heading toward global OSS that's the wrong signal: a foreign contributor opening the tree finds a hardcoded "find-my-native-language" linter, and it implies an ongoing maintenance commitment for something that served a single cleanup pass.

It also **doesn't work as a long-term CI gate**: vi+en code-switching is a *core feature*, so Vietnamese legitimately appears in test data, fixtures, `crawl_markers.py`, the stopword list, TTS samples, and bilingual UI — this PR's own backlog has a long "KEEP" list. A blanket "no Vietnamese" scanner would be noisy and need a constantly-tuned allowlist. Not worth wiring into CI.

**Suggestion:**
- **Remove the script from the PR.** The durable deliverable is `docs/oss-i18n-backlog.md` — it already captured the findings. The script is just the ladder used to build that doc; running it once locally is enough, it doesn't need to be versioned.
- If a *permanent* guard against new VI creeping into prod source is genuinely wanted, make it a **language-neutral English-first lint** (e.g. flag non-ASCII in prod comments/docstrings minus an allowlist) named for its intent — not a VN-specific scanner. Given the legitimate-VI surface above, manual PR review is probably sufficient and this is likely unnecessary.

Net: keep the audit *result* (the doc), drop the *tool*.

omnigentx:
## Re-review of `de9e569` (review follow-ups) — verified

Ran the new tests + probed the regex live. Both code fixes are correct:

- **#3 secret-FP** — confirmed: `[A-Za-z0-9]{32,}` (dropping `-`/`_`) clears the false positives while keeping real secrets:
  ```
  False  uuid (550e8400-e29b-41d4-a716-446655440000)
  False  record_id (mem_01HXYZ7Qpqr8stuvwxyz0123456789ab)
  True   git_sha (40-hex — indistinguishable, only forces approval)
  True   redis://default:Sup3rS3cret@host
  ```
  +2 negative tests. Good.
- **#2 saved-path test** — real cross-task coverage: it resets the ContextVar *before* the `create_task` body runs and still asserts the id is emitted (the `copy_context` snapshot). Exactly the hop the recall test couldn't reach. `test_memory_sensitivity` + the new test → **13/13 pass** locally.

### One new issue introduced by the doc-untracking ⚠️

Untracking `scan_vietnamese.py` + `oss-i18n-backlog.md` addresses my note — but the commit also untracked the **whole design-doc set** (memory-v2-plan, `memory-impl/`, adaptive-rag-spec, crawl-redesign-spec, …). That leaves **~10 shipped source files pointing at docs that no longer exist in the repo**:

```
backend/core/database.py:729:           # ... (docs/agent-memory-adaptive-rag-spec.md)
backend/services/memory/models.py:4:    See docs/agent-memory-adaptive-rag-spec.md §4, §12.1, §14.
backend/services/memory/settings.py:8:  See docs/agent-memory-adaptive-rag-spec.md §18.3 and docs/memory-impl/02.
backend/services/indexing/outbox_service.py:1: ...; docs/memory-impl/03).
backend/services/indexing/memory_index_worker.py:4: ... see docs/memory-impl/03 ...
backend/services/retrieval/contracts.py:2:  See docs/agent-memory-adaptive-rag-spec.md §7.4, §9, §15.
backend/tools/story_server.py:766:        # ... see docs/crawl-redesign-spec.md
backend/services/{memory,indexing,retrieval}/__init__.py:1: "(docs/agent-memory-adaptive-rag-spec.md)"
```

A contributor reading `models.py` is told "See docs/agent-memory-adaptive-rag-spec.md §4" — and that file isn't in the tree. Dangling pointers read worse than no pointer.

Two clean options:
- **(a)** Keep the architecture specs tracked (scrub PII first) — they're an onboarding asset for an OSS repo, and the `See §X` references stay valid. My lean.
- **(b)** Keep them untracked, but in the same commit **strip the inbound `See docs/…` references** from the source comments so nothing points at a missing file.

Either is fine; (c) "leave the dangling refs" is the one to avoid. Also worth noting this untracking is a sizable, somewhat-separate change riding inside a security/PII PR — fine to land, just flagging the scope.

PR #112, opened 2026-06-28
feat(memory): faster reranker default + download-progress UI + per-tool tool results

omnigentx:
## What & why

The agent reranker was the dominant recall cost on the CPU-only deployment and
the chat/memory UX around it had two bugs. This PR makes recall fast by default,
makes a reranker-model switch visible, and fixes per-tool result display.

Investigation notes (measured, on the prod CPU): `docs/rerank-latency-issue.md`.

### 1. Faster reranker default (`feat`)
- Default reranker: **`Qwen/Qwen3-Reranker-0.6B` → `BAAI/bge-reranker-v2-m3`**.
  Qwen3-Reranker is a 600M causal LM measured at **~0.87s/candidate** on CPU →
  ~17s recall for a one-line query. bge-reranker-v2-m3 is a multilingual
  cross-encoder (one forward + head), ~10× cheaper on CPU, handles VI+EN evenly.
- **`rerank_top_k` 20 → 5** — each candidate is a forward pass, so cost scales
  with this; 5 covers the relevant head for personal memory and keeps recall
  well under ~1s.
- Updated dataclass + `_SCHEMA` defaults, `DEFAULT_RERANKER`, and the
  orchestrator/lifespan fallbacks. Qwen3 stays selectable via Settings.

### 2. Realtime reranker-model download progress (`feat`)
Switching the rerank model used to download + load lazily on the FIRST recall —
a multi-second silent hang. Now:
- `reranker.prefetch_and_warm(model, on_progress)` — `snapshot_download` with a
  byte-aggregating tqdm → real % → load → ready (or error). Off the request path.
- `patch_settings` kicks the prefetch in a daemon thread on a `rerank_model`
  change and streams progress to the activity SSE
  (`call_soon_threadsafe`, since broadcast touches non-thread-safe queues).
- `SettingsMemory` shows a progress bar (download % / indeterminate load / ready
  ✓ / error ⚠); recalls during the load keep the old model / fusion order so
  chat never blocks. Full en/vi i18n.

### 3. Per-tool result in the "tools used" panel (`fix`)
A multi-tool turn rendered the FIRST tool's result on every card
(`get_current_time` showed `memory_remember`'s output) because the `tool_done`
SSE carried one batch-level `result_preview`. Now each tool carries its own
result (matched by `tool_id`, positional fallback); FE + the chat-history reload
read per-tool, falling back to the batch field for older rows.

### 4. Global memory SSE routing (`fix`)
`agents.js` dropped agent-less events at the `!agent_name` guard before the
`memory_*` forward gate (so `memory_indexed` and the new
`memory_reranker_loading` never reached the store). Forward agent-less `memory_*`
events before that guard.

## Tests
- Backend: `test_memory_reranker` (prefetch progress phases + dispatch),
  `test_memory_settings` (new defaults), `test_sse_progress` (per-tool mapping +
  positional fallback), `test_memory_retrieval_core` updated. Full memory/sse/
  session sweep **344 passed**.
- Frontend: `toolEvents` (per-tool + fallback), `memory` store
  (`memory_reranker_loading`), `agents` store — **32 passed**. `vite build` clean.

## Manual click-flow
1. Settings → Memory → change Reranker model to a not-yet-cached model → Save →
   a progress bar shows download % → loading → ready ✓; chat stays responsive.
2. Chat a turn that runs ≥2 tools → expand "tools used" → each tool shows ITS
   OWN result.
3. Toggle En/Vi → both panels translate.

### DIFF
--- a/backend/routes/memory_settings.py
+++ b/backend/routes/memory_settings.py
@@ -39,7 +39,9 @@ def _kick_reranker_warm(model_name: str) -> None:
         loop = None
 
     def on_progress(state: str, pct: int) -> None:
-        ev = {"event_type": "reranker_model_loading", "state": state,
+        # ``memory_`` prefix so the frontend's activity-stream router forwards it
+        # to the memory store (agents.js gates forwarding on event_type.startsWith).
+        ev = {"event_type": "memory_reranker_loading", "state": state,
               "progress": pct, "model": model_name}
         if loop is not None:
             loop.call_soon_threadsafe(activity_stream_manager.broadcast, ev)
--- a/backend/routes/memory_settings.py
+++ b/backend/routes/memory_settings.py
@@ -6,7 +6,9 @@ masked on read. Audit history / export still flow through config_service.
 """
 from __future__ import annotations
 
+import asyncio
 import logging
+import threading
 from dataclasses import asdict
 from typing import Any
 
@@ -14,10 +16,38 @@ from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 
 from core.auth import verify_api_key
+from services.activity_stream import activity_stream_manager
 from services.memory.settings import get_memory_settings, update_memory_settings
 
 logger = logging.getLogger("memory_settings_api")
 router = APIRouter(prefix="/api/memory", tags=["memory"])
+
+
+def _kick_reranker_warm(model_name: str) -> None:
+    """After a rerank-model switch, download + warm the new model in a daemon
+    thread and stream progress to the activity SSE — so the UI shows a progress
+    bar instead of the FIRST recall silently hanging on the download/load.
+
+    The warm runs off the event loop (it blocks); progress is marshalled back
+    onto the loop with ``call_soon_threadsafe`` because ``broadcast`` touches
+    asyncio.Queues that are not thread-safe to write from another thread.
+    """
+    from services.retrieval.reranker import prefetch_and_warm
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        loop = None
+
+    def on_progress(state: str, pct: int) -> None:
+        ev = {"event_type": "reranker_model_loading", "state": state,
+              "progress": pct, "model": model_name}
+        if loop is not None:
+            loop.call_soon_threadsafe(activity_stream_manager.broadcast, ev)
+        else:
+            activity_stream_manager.broadcast(ev)
+
+    threading.Thread(target=prefetch_and_warm, args=(model_name, on_progress),
+                     name="reranker-prefetch", daemon=True).start()
 
 
 class MemorySettingsPatch(BaseModel):
@@ -65,4 +95,9 @@ async def patch_settings(patch: MemorySettingsPatch) -> dict[str, Any]:
     except ValueError as exc:
         raise HTTPException(status_code=422, detail=str(exc))
     logger.info("[MEMORY] Settings updated: %s", sorted(updates))
+    # A rerank-model switch would otherwise download + load on the first recall
+    # (a multi-second silent hang). Pre-warm in the background and stream
+    # progress to the UI. Only when the reranker is actually enabled.
+    if "rerank_model" in updates and settings.reranker_enabled:
+        _kick_reranker_warm(settings.rerank_model)
     return asdict(settings)
--- a/backend/server.py
+++ b/backend/server.py
@@ -492,7 +492,7 @@ async def lifespan(app: FastAPI):
             if getattr(_mem_cfg, "reranker_enabled", False):
                 from services.retrieval.reranker import get_shared_reranker
                 get_shared_reranker(
-                    getattr(_mem_cfg, "rerank_model", None) or "Qwen/Qwen3-Reranker-0.6B"
+                    getattr(_mem_cfg, "rerank_model", None) or "BAAI/bge-reranker-v2-m3"
                 ).warm_async()
                 logger.info("[MEMORY] reranker eager-warm kicked off (background)")
     except Exception:
--- a/backend/services/memory/settings.py
+++ b/backend/services/memory/settings.py
@@ -55,10 +55,12 @@ class MemorySettings:
     # evenly. Qwen3 (better gate spread, slower) still selectable. NOTE: an earlier
     # 2026-06-23 note claimed bge "compressed scores near 0 (ungateable)" — that was
     # raw-logit; the CrossEncoder.predict path used here is sigmoid (0..1). VALIDATED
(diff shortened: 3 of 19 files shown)


omnigentx:
## Code Review — PR #112

**+426 / −29 across 18 files** · 4 commits. Four well-scoped changes: faster reranker default, download-progress UI, per-tool result display fix, and global memory SSE routing fix. High-quality work — each commit cleanly explains its *why* and test coverage is solid. Findings below are mostly risk/nuance flags, not blockers.

### Correctness

- ✅ **`call_soon_threadsafe(broadcast, ev)` is sound.** Verified `broadcast` is synchronous (`activity_stream.py:53`, `put_nowait`), so marshalling it onto the loop from the daemon thread is correct — not a coroutine that'd silently never await.

- ⚠️ **Global events are invisible to agent-filtered subscribers.** `broadcast` skips any subscriber whose `_agent_filter` is set when `agent_name != filter` (`activity_stream.py:62`). Global memory events have `agent_name=""`, so a filtered subscriber will never receive `memory_reranker_loading`. Matches existing `memory_indexed` behavior (not a regression), but confirm the Settings page subscribes without a filter or the progress bar silently won't appear.

- ⚠️ **"Recalls during load keep the old model" is only partly true.** `prefetch_and_warm` calls `get_shared_reranker(model_name)` (`reranker.py:270`) which swaps `_SHARED`/`_SHARED_KEY` to the new model *before* `warm()` finishes. Existing orchestrators hold their own `self._reranker` ref (`orchestrator.py:74`) so they keep the old model — good. But any orchestrator constructed *during* the warm window grabs the swapped-but-unwarmed `_SHARED` and pays the load on the request path — the exact hang this PR aims to remove. Window is small (warm runs immediately after the swap, same thread). Consider swapping `_SHARED` only after `warm()` succeeds, or add a comment.

- **Minor — partial id-match in `_attach_per_tool_results`** (`sse_progress.py`): `matched` flips True if *any* id matches; tools whose id doesn't match then get `result_preview = None` with no positional fallback (that only fires when nothing matched). All-or-nothing fallback — unlikely to bite in practice, but worth being aware of.

### Risks

- 🔴 **`rerank_min_score=0.001` was tuned for Qwen3, not bge.** The commit message itself flags this — the old default was chosen *because* "bge compressed every Vietnamese score near 0 (ungateable)". The PR argues the `CrossEncoder.predict` sigmoid path scores differently (0.99/0.00 in a sanity test) so the floor still gates. Plausible but **unvalidated on real VI queries**, and a wrong floor here means silent recall loss (candidates dropped, nothing errors). Given memory recall has already had a "recall=0" production bug, I'd verify this before/right after merge, not "re-validate if recall looks off."

- Confirm the new default model is **cached on the CPU prod** before the first real recall (mitigated by the eager-warm in `lifespan`, `server.py:492` — good that it's there).

### Tests

Strong (344 BE / 32 FE passing). Gaps: no test for the partial id-match case in `_attach_per_tool_results` (only full-match / full-mismatch covered), and no integration test for `rerank_min_score` gating against the real bge model — understandable (needs the model), but it's the highest-risk untested path.

### Recommendation

**Approve with one pre-merge check:** verify `rerank_min_score` isn't silently dropping recall on real Vietnamese queries against the new bge model before this hits CPU prod. Everything else is minor / follow-up.

omnigentx:
Addressed all review findings (commits `bc5e980`, `ec5878f`):

**🔴 `rerank_min_score=0.001` tuned for Qwen3, not bge — validated, not deferred.**
Ran a **27-query VI labelled sweep** on the real store (on-topic direct + indirect, off-topic clear, off-topic keyword-overlap):
| floor | recall(on-topic) | off-clear suppressed |
|--|--|--|
| **0.001** | **1.00** | **1.00** |
| 0.005 | 0.94 | 1.00 |
So 0.001 is the recall/precision sweet spot — no silent recall loss. Codified as a regression test (`tests/test_eval/test_rerank_gate_eval.py`, synthetic VI profile, opt-in) and the settings comment now cites the sweep. Residual: off-topic **keyword-overlap** (e.g. "Angular latest version") still leaks — a scalar floor can't catch topical overlap (needs intent/LLM), same limitation as the old reranker.

**⚠️ Recalls during the warm window could pay the load on the request path.**
Fixed: `prefetch_and_warm` now builds + warms a fresh instance via `get_reranker` and publishes `_SHARED`/`_SHARED_KEY` **only after `warm()` succeeds**. Until then recalls keep the old, already-warm model.

**Minor — partial id-match in `_attach_per_tool_results`.**
Fixed: per-tool positional fallback (any tool whose id didn't match falls back to call-order position), not all-or-nothing.

**⚠️ Global events vs agent-filtered subscribers.**
Confirmed `AppLayout`'s `useRealtimeStream` subscribes with **no `agentFilter`**, so the Settings page receives the agent-less `memory_reranker_loading` event — the progress bar shows. (Same path `memory_indexed` already uses.)

**✅ `call_soon_threadsafe` / eager-warm caching** — confirmed sound; `BAAI/bge-reranker-v2-m3` is already in the prod HF cache and the lifespan eager-warms it.

Full memory/sse/session sweep: **344 backend passed**, **32 frontend passed**, `vite build` clean.

omnigentx:
## Re-review — all findings addressed ✅

Re-checked after `bc5e9805` and `ec5878f2`. Every point from the previous review is resolved:

- 🔴 **`rerank_min_score=0.001` validated** — `ec5878f2` adds `test_rerank_gate_eval.py` (opt-in `memory_eval`, loads the **real** bge cross-encoder, not mocked): 12 on-topic + 5 off-topic labelled VI queries assert recall is kept and off-topic dropped. 27-query sweep cited inline (0.001 → recall 1.00 / off-suppression 1.00; 0.005 → recall 0.94). The settings comment is now "VALIDATED 2026-06-28" instead of "re-validate later". This was the main blocker — nicely closed, and the floor is now regression-guarded for future reranker swaps.

- ✅ **Warm-window race fixed** (`reranker.py`) — `prefetch_and_warm` now builds a fresh instance via `get_reranker`, warms it, and publishes to `_SHARED`/`_SHARED_KEY` **only after** `warm()` succeeds. Recalls keep the old warmed model until the new one is ready — no load on the request path.

- ✅ **Partial id-match fixed** (`sse_progress.py`) — `_attach_per_tool_results` now does per-tool positional fallback (id match, else call-order position) instead of the all-or-nothing `matched` flag.

- ✅ **Global-event visibility (my one open question) — confirmed a non-issue.** The app-boot activity-stream subscription (`AppLayout.vue:52` → `useRealtimeStream`) passes no `agentFilter`, so the backend subscribes with `agent_filter=None` and agent-less `memory_reranker_loading` events are delivered. The progress bar will show.

LGTM — approve. 🤖

PR #113, opened 2026-06-28
feat: Telegram & Zalo messaging gateways (images + slash commands)

omnigentx:
## Messaging Gateways — Telegram & Zalo

Lets users chat with a Jarvis agent from **Telegram** and **Zalo**. Long-polling Bot APIs (no public domain/webhook needed); inbound text + images; in-chat slash commands.

### What's included
- **Gateway abstraction** — `BaseGateway` (allow-list → typing → dispatch → reply → error) + `BotApiGateway` (shared Telegram-style HTTP transport) + thin `Telegram`/`Zalo` subclasses. Add a platform = ~20 lines + 1 registry line.
- **UI** — Settings → *Messaging Gateways*: enable toggle, bot token, allow-list, answering-agent picker, Test Connection, live status. Config stored in `config_service` (DB, encrypted token); `GatewayManager` subscribes and **live-reloads** (no restart). `GET /api/gateways` (status) + `POST /api/gateways/{platform}/test`.
- **Per-chat session binding** — `gateway_chats` table (SSoT, self-healing): each chat continues the same conversation across restarts.
- **Inbound images** — Telegram `getFile`, Zalo `photo_url` (CDN URL; `image/jpg`→`image/jpeg`), passed as multimodal input. *Requires a vision-capable answering model* — declare unknown/proxy models via a gitignored `.fast-agent/model-overlays/` overlay (see `backend/services/gateways/README.md`).
- **Slash commands** (typed in the chat) — `/new` (`/reset`,`/clear`), `/agent <name>` (per-chat agent), `/help`. `/stop` intentionally omitted (sequential poll loop can't interrupt mid-run).
- **Robustness** — benign 409 handling (Telegram self-overlap), bot-token redaction in errors, diff-reload + orphan-poller guard.
- Plus a small **TeamMonitor** design-system fix (disabled button + dropdown contrast in light theme).

### Testing
- Backend: **40 passed** (unit + integration against real SQLite + e2e poll-loop over mocked httpx), ruff clean.
- Frontend: build ✓, Playwright panel e2e **2 passed** (happy path + responsive).
- Live-verified: Telegram + Zalo read images (combo → gpt-5.5); slash commands work.

### Notes for review
- Branched after the i18n work (#97) + a merge of `main`, so the file diff includes those until #97 lands. **The gateway-specific commits are `cba242c` (feature) and `c62c54f` (UI fix)** — review those.
- ⚠️ **Security:** an empty `allow_from` denies everyone (safe default); `["*"]` allows anyone to drive the agent (which has shell/Gmail/IoT tools) — set it to specific user ids in production.

### DIFF
--- a/backend/services/gateways/telegram.py
+++ b/backend/services/gateways/telegram.py
@@ -114,7 +114,12 @@ class TelegramGateway(BotApiGateway):
                     "content_type": ref["content_type"],
                     "data_b64": base64.b64encode(resp.content).decode("ascii"),
                 })
-            except Exception:
-                logger.exception("[telegram] failed to download media file_id=%s",
-                                 ref.get("file_id"))
+            except Exception as e:
+                # The download URL embeds the bot token; raise_for_status() (and
+                # many httpx transport errors) put that full URL in the exception
+                # message. Redact it and do NOT log the raw traceback (exc_info),
+                # which would carry the token too — this is the one media path
+                # that bypasses _call's protection.
+                logger.error("[telegram] failed to download media file_id=%s: %s",
+                             ref.get("file_id"), self._redact(str(e)))
         return out
--- a/backend/services/gateways/telegram.py
+++ b/backend/services/gateways/telegram.py
@@ -1,0 +1,120 @@
+"""Telegram gateway — long-polling via the Bot API.
+
+Telegram's ``getUpdates`` returns an *array* of updates and uses an
+``update_id`` cursor: pass ``offset = max(update_id) + 1`` so each update is
+delivered exactly once and never replayed after a restart hiccup.
+
+Handles text and photos (incl. images sent as documents). A photo's caption
+becomes the message text, and the largest available size is downloaded and
+handed to the agent as multimodal input.
+
+Token: create a bot with @BotFather, put the token in ``fastagent.secrets.yaml``
+under ``gateways.telegram.token`` (see the example file).
+"""
+from __future__ import annotations
+
+import base64
+import logging
+from typing import List, Optional
+
+from .base import InboundMessage
+from .bot_api import BotApiGateway
+
+logger = logging.getLogger("gateways")
+
+# getFile works for files up to 20 MB; skip anything larger than this rather
+# than attempt a download that the Bot API will reject.
+_MAX_MEDIA_BYTES = 20 * 1024 * 1024
+
+
+class TelegramGateway(BotApiGateway):
+    api_base = "https://api.telegram.org"
+    typing_method = "sendChatAction"
+
+    def __init__(self, **kwargs) -> None:
+        super().__init__(name="telegram", **kwargs)
+        self._offset = 0  # update_id cursor; 0 means "from the beginning"
+
+    async def _poll(self) -> List[InboundMessage]:
+        updates = await self._call(
+            "getUpdates",
+            {"offset": self._offset, "timeout": 30, "allowed_updates": ["message"]},
+        )
+        messages: List[InboundMessage] = []
+        for upd in updates or []:
+            # Always advance the cursor, even for updates we skip (stickers,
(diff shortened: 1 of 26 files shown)


backend/services/gateways/telegram.py
```diff
@@ -0,0 +1,120 @@
...
+                continue
+            try:
+                info = await self._call("getFile", {"file_id": ref["file_id"]})
+                file_path = (info or {}).get("file_path")
+                if not file_path:
+                    continue
+                # File download uses a DIFFERENT path: /file/bot<token>/<file_path>
+                url = f"{self.api_base}/file/bot{self._token}/{file_path}"
+                resp = await self._client.get(url)
+                resp.raise_for_status()
```
omnigentx, line 111: **Bot token leaks into logs on a failed media download.**

`fetch_media` builds `url = f"{self.api_base}/file/bot{self._token}/{file_path}"` and calls `resp.raise_for_status()`. On any 4xx/5xx (or many httpx transport errors), the raised `HTTPStatusError` message embeds the full token-bearing URL (`... for url 'https://api.telegram.org/file/bot<TOKEN>/...'`), and the `except Exception: logger.exception(...)` below writes that traceback to the logs — leaking the bot token in plaintext.

Every other error path in this PR goes through `_call` (which never calls `raise_for_status`) or `_redact`/`_safe`. This raw `GET` + `raise_for_status` is the single path that bypasses that protection, directly defeating the PR's stated "bot-token redaction in errors."

**Fix:** redact the token via `_redact(...)` before logging, or catch and strip it from the exception string (log only `ref["file_id"]`).

backend/services/gateways/zalo.py
```diff
@@ -0,0 +1,114 @@
...
+        """Download Zalo images directly from their CDN URL → base64."""
+        out: List[dict] = []
+        for ref in media_refs:
+            url = ref.get("url")
+            if not url:
+                continue
+            try:
+                resp = await self._client.get(url)
+                resp.raise_for_status()
+                if len(resp.content) > _MAX_MEDIA_BYTES:
```
omnigentx: **Media size cap is checked after the full body is buffered.**

`resp = await self._client.get(url)` downloads the entire image into memory, then `if len(resp.content) > _MAX_MEDIA_BYTES:` discards it — so the cap never *prevents* the download, it only drops the result after the fact. Telegram guards correctly with a pre-download `file_size` check (`telegram.py:100`); Zalo does not, so an authorized sender can force the bot to buffer an arbitrarily large CDN response before it's dropped.

Lower severity (only reachable post allow-list gate), but worth a `Content-Length`/streaming pre-check to match Telegram's behavior and make the cap actually bound the work.

omnigentx (commented):
Code review — 1 security finding + 1 robustness finding on the gateway media-download paths. Both name the exact line and a concrete fix. Nice, well-factored feature overall.

backend/services/gateways/telegram.py
omnigentx: Fixed in 42f83aa. The download URL embeds the token, and `raise_for_status()` put it in the `HTTPStatusError` message — the old `logger.exception(...)` then wrote that traceback verbatim. Now the except redacts via `self._redact(str(e))` and logs the message only (no `exc_info`), so the token never reaches the log. Test `test_telegram_media_download_error_does_not_leak_token` asserts the token is absent from `caplog` on a 404 download (and the `file_id` is still logged for debugging).

backend/services/gateways/zalo.py
omnigentx: Fixed in 42f83aa. `fetch_media` now streams: it rejects on the declared `Content-Length` up front, and aborts mid-read if the body crosses the cap anyway (header missing/under-reported) — so the cap actually bounds the buffered work instead of dropping a fully-downloaded body. Test `test_zalo_fetch_media_rejects_oversized_before_buffering` covers the Content-Length rejection.

PR #114, opened 2026-06-29
feat(memory): runtime model prefetch + startup verify (drop Docker bake)

omnigentx:
### Problem
The reranker/embedding models were **baked into `Dockerfile.base`** (a fixed `hf download` list). That list is a *second source of truth* that silently drifts from the config default: v2.10.0 changed the default reranker to `bge-reranker-v2-m3`, but the base still baked the old `Qwen3-Reranker-0.6B` — so the new default **was never on the server**. Confirmed live: the prod container's HF cache had only `bge-m3`, `Qwen3-Embedding-0.6B`, `Qwen3-Reranker-0.6B`. On first recall it would download ~GB on the request path, and re-download **every deploy** (the cache lived in the ephemeral container layer, not a volume).

### Fix — derive the model set from ONE source (live settings), download at runtime, persist on a volume
- **`services/memory/model_prefetch.ensure_models_warm(settings)`** — at startup, download + warm the **configured** embedding + reranker, each in a daemon thread with byte-level progress streamed to the activity SSE (the Settings UI already renders this). Non-blocking: the server is ready immediately; recall degrades gracefully (keeps fusion order / skips rerank) until a model is ready.
- **`server.py` lifespan** now calls it (was reranker-only `warm_async`, no progress).
- **`routes/memory_settings.py`** shares the SSE-progress builder and now also pre-warms the **embedder** on an `embedding_model` change (was reranker-only) — closes the gap where switching the embedder silently hung the first recall.
- **`docker-compose.yaml`** — new **`jarvis-models`** named volume mounted at `/root/.cache/huggingface`, so a runtime download survives `--force-recreate` instead of re-downloading every deploy. (Named volume → `up -V` does **not** wipe it.)
- **`Dockerfile.base` 1.4.0 → 1.5.0** — remove the model bake; the base is system-only again (~5GB smaller). **`Dockerfile`** pins `FROM …base:1.5.0`.

### Tests
- `34 passed` (new `test_model_prefetch.py` + existing reranker/memory-settings suites), ruff clean.
- `ensure_models_warm` kicks both models from config and honors the `enabled` / `reranker_enabled` flags; `prefetch_embedding` emits `downloading → loading → ready` and `error`.

### Deploy steps (manual prerequisite — base image)
1. **Build `omnigentx/jarvis-backend-base:1.5.0`** (system-only). Already built locally on the deploy server (no Docker Hub push needed — `docker compose build` uses the local base, no `--pull`).
2. Merge → CD runs `docker compose build` (app `FROM base:1.5.0`) + `up -d --force-recreate -V`.
3. First startup downloads `bge-reranker-v2-m3` + the embedder into `jarvis-models` (progress in Settings); it persists for subsequent deploys.

### DIFF
--- a/backend/routes/memory_settings.py
+++ b/backend/routes/memory_settings.py
@@ -16,40 +16,41 @@ from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 
 from core.auth import verify_api_key
-from services.activity_stream import activity_stream_manager
 from services.memory.settings import get_memory_settings, update_memory_settings
 
 logger = logging.getLogger("memory_settings_api")
 router = APIRouter(prefix="/api/memory", tags=["memory"])
 
 
+def _running_loop():
+    try:
+        return asyncio.get_running_loop()
+    except RuntimeError:
+        return None
+
+
 def _kick_reranker_warm(model_name: str) -> None:
     """After a rerank-model switch, download + warm the new model in a daemon
     thread and stream progress to the activity SSE — so the UI shows a progress
     bar instead of the FIRST recall silently hanging on the download/load.
-
-    The warm runs off the event loop (it blocks); progress is marshalled back
-    onto the loop with ``call_soon_threadsafe`` because ``broadcast`` touches
-    asyncio.Queues that are not thread-safe to write from another thread.
+    Shares the SSE-progress builder + prefetch with the startup path
+    (services.memory.model_prefetch) so both stay in sync.
     """
+    from services.memory.model_prefetch import make_sse_progress
     from services.retrieval.reranker import prefetch_and_warm
-    try:
-        loop = asyncio.get_running_loop()
-    except RuntimeError:
-        loop = None
-
-    def on_progress(state: str, pct: int) -> None:
-        # ``memory_`` prefix so the frontend's activity-stream router forwards it
-        # to the memory store (agents.js gates forwarding on event_type.startsWith).
-        ev = {"event_type": "memory_reranker_loading", "state": state,
-              "progress": pct, "model": model_name}
-        if loop is not None:
-            loop.call_soon_threadsafe(activity_stream_manager.broadcast, ev)
-        else:
-            activity_stream_manager.broadcast(ev)
-
+    on_progress = make_sse_progress("memory_reranker_loading", model_name, _running_loop())
     threading.Thread(target=prefetch_and_warm, args=(model_name, on_progress),
                      name="reranker-prefetch", daemon=True).start()
+
+
+def _kick_embedding_warm(model_name: str, revision: str) -> None:
+    """Same as :func:`_kick_reranker_warm`, for a newly-selected embedding model
+    — so switching the embedder also downloads/warms with progress instead of
+    hanging the first index write/recall."""
+    from services.memory.model_prefetch import make_sse_progress, prefetch_embedding
+    on_progress = make_sse_progress("memory_embedding_loading", model_name, _running_loop())
+    threading.Thread(target=prefetch_embedding, args=(model_name, revision, on_progress),
+                     name="embedding-prefetch", daemon=True).start()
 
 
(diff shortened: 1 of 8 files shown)


backend/routes/memory_settings.py
```diff
@@ -97,9 +98,12 @@ async def patch_settings(patch: MemorySettingsPatch) -> dict[str, Any]:
     except ValueError as exc:
         raise HTTPException(status_code=422, detail=str(exc))
     logger.info("[MEMORY] Settings updated: %s", sorted(updates))
-    # A rerank-model switch would otherwise download + load on the first recall
-    # (a multi-second silent hang). Pre-warm in the background and stream
-    # progress to the UI. Only when the reranker is actually enabled.
+    # A model switch would otherwise download + load on the first recall (a
+    # multi-second silent hang). Pre-warm in the background and stream progress
+    # to the UI — for the reranker (when enabled) AND the embedder.
     if "rerank_model" in updates and settings.reranker_enabled:
```
omnigentx: **Enabling the reranker doesn't pre-warm it — leaves exactly the cold-load hang this feature targets.**

The guard is `if "rerank_model" in updates and settings.reranker_enabled:`. A user toggling the reranker on sends `{reranker_enabled: true}` → `updates = {"reranker_enabled"}`, so `"rerank_model"` is NOT in `updates` and `_kick_reranker_warm` never fires. The first recall after enabling then cold-loads the 0.6B causal-LM reranker on the request path (~65s on CPU, per this feature's own history). The PR closes this gap for the embedder but re-exposes it for the reranker-enable case.

**Fix:** also warm when `reranker_enabled` flips on, e.g.
```python
if settings.reranker_enabled and ("rerank_model" in updates or "reranker_enabled" in updates):
    _kick_reranker_warm(settings.rerank_model)
```

backend/routes/memory_settings.py
```diff
@@ -97,9 +98,12 @@ async def patch_settings(patch: MemorySettingsPatch) -> dict[str, Any]:
...
         _kick_reranker_warm(settings.rerank_model)
+    if "embedding_model" in updates and settings.embedding_model:
```
omnigentx: **A `embedding_revision`-only change doesn't pre-warm the embedder.**

`embedding_revision` is independently patchable (`MemorySettingsPatch`), and the shared embedding-provider key is `(model_name, revision)`. So pinning a new revision without touching `embedding_model` produces a new provider that downloads/loads on the first recall — the same silent hang the embedder warm was just added to prevent. The guard here only checks `"embedding_model" in updates`.

Lower severity than the reranker case (revision pinning is an advanced action), but it's the same class of gap.

**Fix:** kick the embedder warm when either field changes — `if ("embedding_model" in updates or "embedding_revision" in updates) and settings.embedding_model:`.

omnigentx (commented):
Code review — solid PR. Verified the load-bearing claims (HF_HOME matches the volume mount, reranker/embedder both publish to the shared singleton, settings fields + provider methods exist). 2 findings on the config-change warm-trigger: it doesn't cover all the state transitions that change the active model, so it re-exposes the exact first-recall cold-load hang this feature exists to remove.

backend/routes/memory_settings.py
omnigentx: Fixed in b8df67b. The reranker now warms when it's on AND `("rerank_model" in updates or "reranker_enabled" in updates)` — so the enable-only patch (`{reranker_enabled: true}`) fires it. `updates` is the submitted patch keys (patch_settings:95), so this catches the enable case. Test `test_patch_warms_on_reranker_enable_and_revision_change` asserts it.

backend/routes/memory_settings.py
omnigentx: Fixed in b8df67b. The embedder now warms when `("embedding_model" in updates or "embedding_revision" in updates)`, so a revision-only pin also pre-warms (the new provider key `(model, revision)`). Covered by the same test (revision-only patch → embedder warm fires).

PR #11, opened 2026-05-09
fix(setup): wizard mints session cookie so exit doesn't pop AuthGate

omnigentx:
## Symptom

User on staging finishes the Setup Wizard, navigates to `/agents`, and immediately sees the AuthGate modal:

> Authentication required — Your session has expired. Please log in again.

…even though they typed their key seconds ago. Reported here: comment in this thread + screenshot.

## Root cause

`POST /api/setup/auth` is the wizard's bootstrap endpoint — it **sets** the API key on the backend but never mints a session cookie. The dashboard's `auth.probe()` at the next non-bare route hits `/api/auth/whoami`, sees no cookie, transitions the auth store to `UNAUTHENTICATED`, and the AuthGate opens.

The "session expired" copy is misleading on top of it (the user never had a session in the first place), but the actual UX fix is to mint a cookie at the right moment — right after the user proves they own the key.

This gap was introduced when PR #10 added the auth store but did not patch the setup-store path that completes the auth bootstrap.

## Fix

`stores/setup.js::submitAuth` — after `POST /api/setup/auth` succeeds, lazy-import the auth store and call `.login(apiKey)`. Best-effort: a failure here is non-fatal (older backend, network blip), and the AuthGate now serves as the recovery path.

Lazy import on the auth store keeps the setup store independent of Pinia init order.

`setup_wizard_fresh.yaml` E2E fixture gains a `POST /api/auth/login` mock with the same shape the real backend returns, so the existing "no unexpected requests" invariant still passes.

## Test plan

- [x] Dashboard E2E: 55 / 55 passing
- [x] Dashboard unit: 36 / 36 passing
- [x] `vite build` green
- [ ] Smoke staging: complete Setup Wizard end-to-end, navigate to `/agents`, AuthGate must NOT open

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #115, opened 2026-06-29
feat(memory): per-agent memory — every agent gets its OWN isolated silo

omnigentx:
## What

Every agent now has its **own** durable memory silo, instead of only Jarvis having memory tools while the pooled `memory_server` resolved its owner from a static `MEMORY_AGENT_NAME="Jarvis"` (so other agents had no tool, or wrote into Jarvis's silo).

**One mechanism — caller-identity** (needs submodule fast-agent#11):
- fast-agent stamps a trusted `_meta.caller_agent` (the calling agent's own name) on every tool call; spawned agents are named with their real identity (not `"child"`). Same signal identifies in-process and spawned callers.
- `memory_server._owner()` = `caller_agent`, full stop — no env/static fallback chain; a missing identity FAILS the op rather than mis-scoping. (`MEMORY_AGENT_NAME` removed from config.)
- `memory_server` attached to ALL agents via a shared `_MEMORY_SERVERS` (standalone) + the team templates (agile/research) for spawned; shared `_MEMORY_PROMPT`.

**Lifecycle cleanup — one reused `purge_agent_memory`** (SQLite tables + FTS mirror + `LadybugStore.purge_owner`):
- manual delete: `routes/agents` delete_agent / delete_team.
- oneshot completion: `spawn_progress_bridge` purges on `lifecycle_after_cleanup` (emitted only for a removed oneshot), so throwaway oneshot agents leave no dead silos.

## Verified live (real API)

- IoTAgent (in-process) and a spawned MemTestCarol each wrote to their **own** silo (owner = the agent, never `"child"`/Jarvis); recall returns their own facts.
- Deleting a registered agent purged its memory; a seeded silo for a oneshot was auto-purged on completion.
- Full services+eval suite passes; unit/integration in `tests/test_services/test_memory_owner_identity.py` + `test_spawn_progress_bridge.py`.

## Depends on

- omnigentx/fast-agent#11 (must merge first; this PR bumps the submodule pointer to it).

### DIFF
--- a/backend/agent.py
+++ b/backend/agent.py
@@ -86,11 +86,24 @@ CORE_SKILLS = get_skills("user-context")
 
 # --- Static Agents (require TagRelayAgent or specific hooks) ---
 
+# Every agent gets its OWN durable memory (per-agent silo, isolated by the
+# caller-identity fast-agent stamps on each tool call — see tools/memory_server.py).
+# Spread into each agent's `servers` so a new agent opts in with one token.
+_MEMORY_SERVERS = ["memory_server"]
+# One shared memory instruction for every memory-capable agent (single source).
+_MEMORY_PROMPT = (
+    "\n\nYou have your OWN private long-term memory (separate from other agents). "
+    "Call memory_search BEFORE asking the user something they may have already told you; "
+    "call memory_remember to store durable facts/preferences when the user says to remember "
+    "or states a lasting fact. Never write memory to files."
+)
+
+
 @fast.agent(
     name="PersonalAgent",
-    instruction="You are a personal assistant.\n\n{{agentSkills}}",
+    instruction="You are a personal assistant." + _MEMORY_PROMPT + "\n\n{{agentSkills}}",
     skills=CORE_SKILLS + get_skills("personal-assistant", "cron-management", "scrape-web"),
-    servers=["serpapi", "time-service", "gmail", "calendar", "cron-server", "scrapling-server"],
+    servers=["serpapi", "time-service", "gmail", "calendar", "cron-server", "scrapling-server", *_MEMORY_SERVERS],
     request_params=RequestParams(parallel_tool_calls=True),
 )
 async def personal_agent(prompt: str):
@@ -98,9 +111,9 @@ async def personal_agent(prompt: str):
 
 @fast.agent(
     name="IoTAgent",
-    instruction="You are an IoT specialist.\n\n{{agentSkills}}",
+    instruction="You are an IoT specialist." + _MEMORY_PROMPT + "\n\n{{agentSkills}}",
     skills=CORE_SKILLS + get_skills("iot-control"),
-    servers=["iot-control", "time-service", "gmail"],
+    servers=["iot-control", "time-service", "gmail", *_MEMORY_SERVERS],
     tools={"time-service": ["get_current_time", "wait_for_seconds"]},
     request_params=RequestParams(parallel_tool_calls=True),
 )
@@ -110,9 +123,9 @@ async def iot_agent(prompt: str):
 @fast.custom(
     TagRelayAgent,
     name="MusicAgent",
-    instruction="You are a music specialist.\n\n{{agentSkills}}",
+    instruction="You are a music specialist." + _MEMORY_PROMPT + "\n\n{{agentSkills}}",
     skills=CORE_SKILLS + get_skills("music-playback"),
-    servers=["media-server"],
+    servers=["media-server", *_MEMORY_SERVERS],
     tools={"media-server": ["search_youtube"]}
 )
 async def music_agent(prompt: str):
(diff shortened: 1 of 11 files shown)


omnigentx (commented):
## Review — per-agent memory silo (consumer of fast-agent #11/#12)

Read the full diff (12 files, +346/−52) and ran the suite locally: **`31 passed`** (`test_memory_owner_identity.py` + `test_spawn_progress_bridge.py`). This is a well-designed PR — single-source identity, fail-closed, real integration coverage. **No blockers.** Findings below are hardening/nits.

### ✅ Verified
- **Single source of truth for identity, fail-closed.** `_owner(ctx)` resolves *only* the per-call `caller_agent`; no env/static fallback. A missing identity returns `""` → the tool returns `"no bound agent identity"` rather than mis-scoping a write. This removes the old `TEAM_MY_NAME`/`MEMORY_AGENT_NAME` dual-source — exactly the kind of "try X, fall back to Y for the same decision" that silently diverges.
- **Per-call, not per-connection** — correctness of pooling a single `memory_server` subprocess across in-process agents rests on `caller_agent` being on each tool-call's `_meta`; it is. Concurrent calls from different agents can't cross-talk.
- **Migration safe.** Main agent is `name="Jarvis"` (agent.py:71, 303) → `caller_agent="Jarvis"` → existing `owner="Jarvis"` memory stays reachable after `MEMORY_AGENT_NAME` removal.
- **Real integration test.** `test_caller_agent_meta_flows_through_fastmcp` drives a real FastMCP connected client (`create_connected_server_and_client_session`), not a mock of the subsystem under test — satisfies "Happy Path FIRST".
- **Oneshot-purge collision invariant holds.** Traced every oneshot spawn path: all go through `ensure_unique_agent_name()` / `_generate_unique_agent_name()`, which check the union of live registry + team sessions + `agent_definitions` (persistent agents). `lifecycle_after_cleanup` is emitted only for `lifecycle == "oneshot"` (explicit enum check, after registry removal). So a oneshot can't share a name with "Jarvis"/"IoTAgent"/etc., and its completion can't wipe a persistent silo.
- **Submodule pointer is correct.** `backend/fast-agent` → `0ed25a9d` (the #12 merge), so it includes the `role→agent_name` NameError fix. If it pointed only at #11, every spawned agent's first LLM turn would `NameError` and the spawned-agent memory path (the headline feature) would be dead. Good catch to include #12.

### 🟠 1. Destructive purge trusts an invariant enforced in another module
`_maybe_purge_oneshot_memory` deletes a silo by name, relying on spawn-time uniqueness (in the fast-agent submodule's `ensure_unique_agent_name`) to guarantee a oneshot never collides with a persistent agent. That holds today. But a *destructive* op infers "safe to delete" from the `lifecycle` field instead of probing the authoritative source (the `agent_definitions` registry). The fast-agent history shows new creation paths keep getting added ("enforce unique identity across **all** creation paths" exists precisely because one was missed before). If a future path bypasses the uniqueness gate, this silently wipes a persistent silo.

Cheap fail-safe: in `_maybe_purge_oneshot_memory`, skip the purge if the name exists in `agent_definitions`. It makes the delete fail-safe regardless of upstream invariants — worthwhile insurance for an irreversible op.

### 🟠 2. PR body understates the submodule bump
Body says it "bumps the submodule pointer to #11", but the actual pointer is `0ed25a9d` = the **#12** merge. The pointer is right (you *need* #12); just update the description so a future reader doesn't think #12 is missing.

### 🟢 3. (Minor) `_caller_from_ctx` swallows the meta-access exception silently
The `except Exception → return ""` branch is correct (fails closed, and the tool surfaces the error downstream), but a `logger.debug` there would help diagnose a transport that doesn't expose `request_context.meta`.

### 🟢 4. (Confidence note, not a bug) `ctx: Context = None`
FastMCP excludes `Context`-typed params from the LLM-facing tool schema, so `caller_agent` can't be influenced by the model — consistent with the design and the e2e flow test.

### Verdict
Approve. Recommend #1 (one-line fail-safe) before merge given the op is irreversible; #2 is a description fix; #3/#4 optional.

omnigentx:
Addressed the review (force-pushed `aadbd39`):

**🟠 1 — Destructive purge now fail-safe (the important one).** `_maybe_purge_oneshot_memory` no longer trusts spawn-time uniqueness for an irreversible delete — it probes the authoritative sources directly and **refuses to purge a persistent agent's silo**: skip if the name is a live in-process agent (`state.agent_app._agents` — covers Jarvis/IoTAgent/etc.) OR has an `agent_definitions` row (dynamic persistent). If the lookup can't verify (raises), it **fails safe = does not delete**. Tests: `test_oneshot_purge_fail_safe_refuses_persistent_agent` (a oneshot event for "Jarvis" → not purged; a real throwaway → purged) + `test_oneshot_purge_fail_safe_skips_when_unverifiable`.

**🟠 2 — Submodule pointer clarification.** Confirmed: `backend/fast-agent` → `0ed25a9d` = the **#12** merge (carries the `role→agent_name` NameError fix), not #11. The PR body wording was misleading; the pointer itself is correct. (Body edit hit a GitHub projects-classic deprecation error; noting it here so the record is clear.)

**🟢 3 — `_caller_from_ctx` now logs on meta-access failure.** Added a `logger.debug(..., exc_info=True)` in the `except` branch so a transport that doesn't expose `request_context.meta` is diagnosable; still fails closed.

🟢 4 — noted, no change (Context param is excluded from the LLM-facing schema, by design).

Suite still green (33 incl. the 2 new fail-safe tests).

PR #116, opened 2026-06-29
fix(deploy): widen healthcheck window for cold model-cache first deploy

omnigentx:
## Problem

The `v2.11.0` self-hosted deploy false-failed: `auth/probe=000` for the full 360s window, then exit 1 — but the container never crashed (it was still booting).

**Root cause:** the runtime-model-prefetch change downloads BOTH the embedding (~1.2GB) and the reranker (~2.2GB) ≈ **3.4GB** from HF Hub at boot (unauthenticated → rate-limited) into a **cold** `jarvis-models` volume, then re-embeds the store. The lifespan blocks at `async with fast.run()` (uvicorn doesn't serve the probe until startup completes) well past the **360s** window, which was sized for 1.2GB. The models persist in the volume, so **only the first cold deploy** pays this; later deploys hit the cache and come up in seconds.

## Fix

- **Window 360s → 900s** to cover the cold-cache 3.4GB download + re-embed.
- **NEW stall guard:** if the backend stops logging for **180s**, fail fast — a genuine hang produces no output, whereas a slow download streams continuous progress lines. So the long ceiling never masks a real hang.
- Keeps the existing crashed-container fast-fail.
- TODO in-comment: set an `HF_TOKEN` secret (lifts the unauthenticated rate limit) and/or pre-seed the `jarvis-models` volume to shrink first-deploy time.

Not related to per-agent-memory (#115) — this is a main/#114 deploy-window issue. A retry of the failed deploy will likely already pass now that the models finished downloading into the volume in the background.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -111,38 +111,57 @@ jobs:
           # can verify the process is up without needing a bearer token.
           # /api/setup/status would work pre-master-key (dev mode) but starts
           # returning 401 once core_auth.JARVIS_API_KEY is populated.
-          # 360s window: a FIRST deploy with a new embedding/reranker model
-          # downloads ~1.2GB from HF Hub AND re-embeds the store at boot, which
-          # blocks startup well past the old 120s (the v2.7.0 Qwen swap false-
-          # failed here while the backend was in fact coming up). A crashed
-          # container fails fast below, so the long window only ever waits on a
-          # genuinely-slow-but-healthy boot.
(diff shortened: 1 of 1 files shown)


omnigentx:
Closing — replacing the deploy.yml timeout band-aid with the proper fix (decouple the liveness probe from model/agent init so the server is 'up' immediately and models download in the background with the existing SSE progress UX). That fix lands in #115 per request to keep it one PR.

PR #117, opened 2026-06-29
fix(startup): serve liveness probe immediately — background the agent runtime

omnigentx:
## Problem

The `v2.11.0` deploy false-failed: `auth/probe=000` for the entire health window, then exit 1 — but the container never crashed. Root cause: the lifespan blocks at `async with fast.run()` (eager MCP/agent connect), and **uvicorn serves no request until lifespan startup returns**. On a cold first deploy that also downloads ~3.4GB of models at runtime, startup exceeds the fixed window even though the backend is coming up fine.

Bumping the window (the closed #116) is a band-aid that doesn't scale — a heavier model means bumping again.

## Fix — decouple liveness from heavy init

Bring the FastAgent runtime up in a **background task** and `yield` immediately, so uvicorn serves the liveness probe (which never needed `agent_app`) in seconds while agents + MCP servers + the model cache warm up (the existing SSE progress bar already shows download status). **The probe is now independent of model size** → no more timeout chasing.

### Safety
- `agent_app`-dependent routes return **503 "starting up"** instead of crashing until the background runtime sets `state.agent_app` (chat / chat_stream guarded; most other routes already check it).
- `reload_task` / `gateway_manager` (created inside the background runtime) are read back through `state` for shutdown.
- Background-task exceptions are logged via a done-callback.

### Tradeoff
The CD probe is now a **liveness** check (process up), not **readiness** (agents ready). A genuinely broken agent init would deploy "up" but serve 503 — surfaced by the readiness gate + monitoring rather than a deploy failure. A dedicated `/ready` endpoint could gate that separately later.

## Tests
`tests/test_routes/test_startup_readiness.py`: probe answers without `agent_app`; chat/chat_stream return 503 until ready. Chat e2e + setup route suites pass (36).

## ⚠️ Risk
gitnexus: **HIGH** — touches `lifespan` + `chat`/`chat_stream`. **Needs a staging deploy** to verify real cold first-deploy boot before prod. Separate from per-agent-memory (#115) per request.

### DIFF
--- a/backend/routes/chat.py
+++ b/backend/routes/chat.py
@@ -8,7 +8,7 @@ import asyncio
 import logging
 import time as _time
 
-from fastapi import APIRouter, Request, Depends
+from fastapi import APIRouter, Request, Depends, HTTPException
 from pydantic import BaseModel
 
 from core.auth import verify_api_key
@@ -292,6 +292,9 @@ def _prepare_audio_url(book_id, tts_text, base_url=""):
 @router.post("/chat", response_model=ChatResponse)
 async def chat(request: ChatRequest, raw_request: Request = None, _=Depends(verify_api_key)):
     from services.shared_state import agent_app
+    if agent_app is None:
+        raise HTTPException(status_code=503,
+                            detail="Backend is still starting up (agents not ready). Retry shortly.")
     from services.sse_progress import current_run_id
     _t0 = _time.time()
     _msg_preview = request.message[:80] + "..." if len(request.message) > 80 else request.message
@@ -347,7 +350,10 @@ async def chat_stream(raw_request: Request, _=Depends(verify_api_key)):
     """
     from sse_starlette.sse import EventSourceResponse
     from services.shared_state import agent_app
-    
+    if agent_app is None:
+        raise HTTPException(status_code=503,
+                            detail="Backend is still starting up (agents not ready). Retry shortly.")
+
     _t0 = _time.time()
     
     # Parse request: multipart/form-data or JSON
--- a/backend/routes/setup.py
+++ b/backend/routes/setup.py
@@ -199,6 +199,30 @@ async def auth_probe():
     return AuthProbe(configured=bool(core_auth.JARVIS_API_KEY))
 
 
+class ReadinessProbe(BaseModel):
+    """Readiness (vs liveness). 200 only once the background FastAgent runtime has
+    set ``state.agent_app`` (agents/MCP up); 503 until then."""
+
+    ready: bool
+    agents: int = 0
+
+
+@router.get("/ready", response_model=ReadinessProbe)
+async def readiness_probe(response: Response):
+    """READINESS probe — distinct from the liveness ``/auth/probe``. Liveness answers
+    as soon as the process is up (so a slow first-deploy model download can't false-
+    fail CD); readiness answers 200 only once agents are actually initialized.
+    Unauthenticated so CD / monitoring can alert on a backend that stays live but
+    never becomes ready — i.e. a genuine init failure surfaces as a persistent 503
+    here instead of a silently-green deploy."""
+    import services.shared_state as _state
+    app = _state.agent_app
+    if app is None:
+        response.status_code = 503
+        return ReadinessProbe(ready=False, agents=0)
+    return ReadinessProbe(ready=True, agents=len(getattr(app, "_agents", {}) or {}))
+
+
 @router.post("/auth", response_model=SetupStatus)
 async def setup_auth(payload: AuthStep, request: Request, response: Response):
     """Step 1 — choose, adopt, or generate the master key.
--- a/backend/server.py
+++ b/backend/server.py
@@ -565,284 +565,322 @@ async def lifespan(app: FastAPI):
     except Exception as e:
         logger.warning("MCP catalog boot sync failed: %s", e, exc_info=True)
 
-    async with fast.run() as agent:
-        state.agent_app = agent
-        logger.info("FastAgent initialized.")
+    # Bring the FastAgent runtime up in the BACKGROUND so the lifespan yields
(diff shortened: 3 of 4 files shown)


omnigentx (commented):
## Review — serve liveness immediately, background the agent runtime

Read the full diff (`server.py` lifespan restructure, `chat.py` guards, new readiness test) and ran the suite: **`3 passed`**. The core mechanism is correct and the warmup window is guarded more thoroughly than the description claims. **No code blockers** — the real gate is the staging deploy the PR already calls for.

### ✅ Verified correct
- **No SSoT divergence on `agent_app`.** Write is `state.agent_app = agent` where `state` is `import services.shared_state as state` (server.py:113, 580); chat reads `from services.shared_state import agent_app` (re-imported per call → sees the live value). Same module attribute, symmetric read/write. Default `agent_app = None` (shared_state.py:37) so the import can't fail before first write.
- **`fast.run()` stays open for the app's life.** `_agent_runtime` ends on `await _agent_shutdown.wait()` *inside* the `async with` (server.py:842), so `agent_app` isn't torn down right after setup. Shutdown signals it (`_agent_shutdown.set()`) with a bounded `wait_for(..., 30)` + cancel fallback (858-862). Clean.
- **Warmup-window guards are comprehensive**, not just chat:
  - `chat` / `chat_stream` → 503 (new).
  - `inject` → already 503 (inject.py:327).
  - `routes/agents.py` (359/414/455/507), `ws_voice.py` (706) → guarded.
  - `cron_scheduler` skips `agent_turn` jobs while the ref is null (cron_scheduler.py:545) — important, since `cron_task` starts at line 436 *before* the background runtime sets agent refs (line 648). So a cron firing mid-warmup is handled, not a crash.
  - null-safe readers (`fast_extractor.py:266`, `spawn_progress_bridge.py:582`) use `getattr(..., default)`.
- **Background failure is logged** via the done-callback (844-849).
- **Bonus fix bundled:** moving the KG backfill to *after* `state.agent_app = agent` (625-631) repairs a real latent no-op (extractor LLM was unresolved before). Good catch.

### 🟠 1. Liveness-only probe → a broken init deploys "green" but serves 503 forever
This is the PR's stated tradeoff, but it's the main risk worth hardening. If `fast.run()` raises or hangs during init, the done-callback logs it, but `state.agent_app` stays `None` permanently → every chat returns 503 while **CD reports the deploy succeeded and the container stays up**. Nothing pages; it's invisible until a user complains. The "silent fallback" is no longer in the route (the 503 is honest) — it's moved to the *deploy signal*, which now passes regardless of agent health.

Recommend landing the `/ready` endpoint sooner than "later", OR at minimum a monitoring alert on `agent_app is None` beyond N minutes post-boot. Otherwise this trades a noisy false-failure for a silent true-failure.

### 🟠 2. MCP admin routes read `agent_app` unguarded → 500 (not 503) during warmup
`mcp.py:78`, `mcp_admin_service.py:814`, `mcp_attachments.py:310`, `mcp_catalog.py:267` read `agent_app` with no `None` check, so they'd raise a 500 if hit in the warmup window. Not catastrophic (admin endpoints, brief window, liveness unaffected), but it contradicts "most other routes already check it" — these don't. Either add the 503 guard for consistency or note them as accepted-500.

### 🟢 3. (Edge, acceptable) Cold-deploy teardown
If a shutdown lands while `fast.run()` is mid multi-GB download (>30s connect), `_agent_shutdown.set()` can't take effect until the task reaches `wait()`; `wait_for` times out → `cancel()` interrupts the download. Handled, just noting the cold-boot-then-redeploy case cancels mid-pull.

### 🟢 4. (Minor) gateway shutdown gap
If `GatewayManager(...)` constructs but `.start()` raises, `state.gateway_manager` is never set (834 runs after 833), so shutdown won't call `.stop()`. Harmless if start failed, but the local `gateway_manager` is then orphaned.

### Verdict
Approve from a code-correctness standpoint. Agree with the self-assessed HIGH risk: the one thing tests can't cover is the actual cold first-deploy boot timing — **verify on staging before prod**. Push #1 (readiness gate / alert) as a fast follow so a genuine init failure isn't silent.

omnigentx:
Addressed the review (force-pushed `881fcf7`):

**🟠 1 — Silent-failure risk → landed `/ready` now (not "later").** New `GET /api/setup/ready` (unauthenticated, in `routes/setup.py`): 200 only once the background runtime sets `state.agent_app`, **503 until then**. So a backend that stays *live* but never becomes *ready* (a genuine init failure) shows as a persistent 503 on `/ready` that CD/monitoring can alert on — the liveness `/auth/probe` stays the fast deploy-gate. Tests: `/ready` 503 until ready, 200 + agent count when ready. (Wiring CD to also soft-probe `/ready` is an ops/deploy.yml follow-up; the endpoint is the mechanism.)

**🟠 2 — MCP admin reads: verified already guarded, no 500.** Re-checked the four spots — each does `agent_app = shared_state.agent_app; if agent_app is None: return <graceful error>`: `mcp.py:79` → `"unknown"`, `mcp_admin_service.py` → `{"ok": False, "error": "agent_app not initialized"}`, `mcp_attachments.py` → `(None, "...not ready")`, `mcp_catalog.py` → `{"ok": False, ...}`. They return graceful errors, not 500s — so "most other routes already check it" does hold for these. No change.

**🟢 4 — Gateway shutdown gap fixed.** `state.gateway_manager` is now set BEFORE `gateway_manager.start()`, so a manager that constructs but fails mid-start is still reachable for shutdown `.stop()` (no orphan).

**🟢 3 — cold-deploy teardown:** accepted as-is (the bounded `wait_for(30)` + cancel correctly interrupts a mid-download teardown; redeploy resumes from the cache volume).

Still green locally (33 incl. the 2 new `/ready` tests). The real gate remains the **staging cold-boot deploy** — unchanged from the PR's self-assessment.

PR #13, opened 2026-05-09
refactor(mcp): built-in MCPs talk to backend over Runtime RPC, drop API-key env

omnigentx:
## Summary

- Built-in MCPs (`approval-server`, `skill_server`, `mcp_admin`) now talk to the backend over the existing `RuntimeRpcServer` Unix socket. No HTTP, no Bearer key. Trust = file-system perms on the socket path (same model already used by `skill_server` and `mcp_admin`).
- `approval-server` is the only one that previously needed migrating; the `JARVIS_API_KEY` / `JARVIS_API_BASE` lines on `skill_server` + `mcp_admin` were vestigial (code already on UDS RPC, not reading those vars).
- New: `RuntimeRpcServer.register(..., timeout=None)` opts out of the 30 s dispatch deadline so `approval.wait` can long-poll until a human resolves.
- **Bundled**: also includes `dashboard/nginx.conf` `/ws/` proxy fix for hands-free voice WebSocket (commit `3c5e0f8`). Bundled rather than separate PR to ship as one CD batch — see commit message for full context. Originally PR #14, now closed and folded in here.

## Why

These two env vars were the last reason built-in MCPs needed an API key. Bearer is fine for external clients (`xiaozhi_integration`), but trusted in-process tools shouldn't be sending HTTP back to themselves — slower than UDS, requires CSRF exemptions, and means a dashboard XSS surface could lift a key intended for subprocess auth. Trust on the UDS socket path is bounded by file-system permissions on the runtime dir, which is the existing model for skill_server/mcp_admin.

## What changed

| File | Change |
|---|---|
| `services/runtime_rpc.py` | `register()` accepts per-method `timeout=`. `None` = unbounded (long-poll). |
| `services/approval_service.py` | In-process pub/sub (`_resolution_waiters`) + async `wait_for_resolution()`. `resolve_approval` signals waiters at end. Single-process by design (Jarvis is a personal AI assistant). |
| `services/approval_rpc_handlers.py` (new) | `approval.create`, `approval.get`, `approval.wait` (last one `timeout=None`). |
| `tools/approval_server.py` | Rewritten — no httpx, no API key. Uses `rpc_call` via `asyncio.to_thread`. Reconnect loop with exponential backoff (2 s → cap 30 s) on socket drop. |
| `tools/runtime_rpc_client.py` | `call(timeout=)` accepts `None` for blocking-mode sockets. |
[...]

PR #119, opened 2026-06-30
feat(compaction): spawned agents self-compact (parity with in-process)

omnigentx:
## Problem

Only **in-process** agents compacted their context (`server.py` attaches `attach_compaction_hooks_to_all` to every in-process agent). **Spawned** team/dynamic agents — which run their own long conversations in a subprocess — were never hooked, so their `message_history` grew **unbounded** → token bloat / context-limit risk on long tasks.

## Fix

Wire the **same** compaction hook into `isolated_runner` after the spawn hook chain (fast-agent #13). The backend is importable in the spawned process (`PYTHONPATH=project_dir`, same as the existing `context_persistence` import), so it reuses the exact in-process hook — `merge_hooks` keeps pause-first ordering and OR-merges `on_context_overflow` → **full parity** (proactive + emergency compaction).

## Verified

- **Integration** (`tests/test_services/test_spawned_compaction.py`): attach merges onto a spawned-style hook chain — both hooks run, ordering preserved, `on_context_overflow` survives, idempotent.
- **Live** in the real spawned-process env: `import services.context_compaction` + attach → `attached=1, sentinel=True, has_overflow=True`.
- Bonus: confirmed `/api/setup/ready` (#117) live — 503 during warmup → `{"ready":true,"agents":8}`.

(Actual firing needs >~30% of the model window; that's `maybe_compact_agent`, already covered by the in-process compaction suite.)

## Depends on
- omnigentx/fast-agent#13 (merge first; this PR's submodule pointer = `f25a7e0c`, re-point to the merge commit before merging here).

### DIFF
--- a/backend/tests/test_services/test_spawned_compaction.py
+++ b/backend/tests/test_services/test_spawned_compaction.py
@@ -1,0 +1,64 @@
+"""Spawned agents self-compact too (parity with in-process).
+
+isolated_runner now calls ``attach_compaction_hooks_to_all`` on the spawned agent
+after installing the spawn hook chain, so a long-running team/dynamic agent compacts
+its own message_history instead of growing it unbounded (only in-process agents did
+before). These tests pin the merge behaviour the spawned path relies on.
+"""
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import services.context_compaction as cc
+from fast_agent.agents.tool_runner import ToolRunnerHooks
+
+
+async def test_compaction_merges_onto_spawned_hook_chain(monkeypatch):
+    """A spawned agent already carries the spawn hook chain (pause/spawn/rtac). The
+    compaction attach MERGES in: both the pre-existing before_llm hook AND the
+    proactive compaction run, and the attach is idempotent."""
+    ran: list[str] = []
+
+    async def existing_before_llm(_r, _m):
+        ran.append("spawn")
+
+    async def fake_compact(*_a, **_k):
+        ran.append("compact")
+
+    monkeypatch.setattr(cc, "maybe_compact_agent", fake_compact)
+
+    agent = SimpleNamespace(
+        name="DevAgent",
+        tool_runner_hooks=ToolRunnerHooks(before_llm_call=existing_before_llm),
+    )
+    app = MagicMock(_agents={"DevAgent": agent})
+
+    assert cc.attach_compaction_hooks_to_all(app) == 1
+    assert getattr(agent, "_jarvis_compaction_hook", False) is True
+
+    runner = SimpleNamespace(_agent=SimpleNamespace(name="DevAgent", message_history=[]))
+    await agent.tool_runner_hooks.before_llm_call(runner, [])
+    assert ran == ["spawn", "compact"]            # spawn first, compaction merged after
+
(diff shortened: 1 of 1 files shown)


omnigentx:
✅ **Live firing verified** (local 8001, real spawned agents, then cleaned up):

Key: a single big fetch gets truncated to a file at >64KB so it never accumulates context — the agent must read in **small chunks, many times**. With `max_context_tokens=10000, compact_at_ratio=0.3` (threshold 3000) a real spawned agent reading a 14KB file 6× fired compaction for real:

- **Oneshot** (`CompactDemo`): 4 events, e.g. `10418→6140` tokens (~41% reduction), all `auto_threshold`/`completed`.
- **Resumable** (`CompactResume`): 3 events (`1880→1564`, `2251→1787`, `7002→5888`), and it stays in the agent list → viewable in the UI at **AgentDetail → Context Versions** tab.

The DB-path fix was load-bearing: before it, the spawned process read compaction config from a stray cwd-relative DB → DEFAULTS → never fired with tuned settings. After it, config + events resolve to the backend DB. (27× real `filesystem__read_text_file` tool calls over ~117s confirm a real LLM agent, not a mock.)

omnigentx (commented):
## Review — verified, safe to merge ✅

This is the thin jarvis-side companion to fast-agent#13: a submodule pointer bump + the backend-side integration test. The actual code (the `isolated_runner` compaction hook and the `isolated_spawner` `JARVIS_DB_PATH` pin) I reviewed in detail on fast-agent#13 and confirmed correct. Here I verified the integration mechanics:

### Submodule pointer — safe
- New pointer `60a40e2` **is an ancestor of fast-agent `origin/main`** — #13 is already merged (merge commit `2d56b7c9`). So this is **not** a floating/dangling pointer; `git submodule update` resolves cleanly on a fresh clone. ✅
- Clean forward bump `0ed25a9 → 60a40e2`, no history rewrite. The only two commits introduced are exactly the expected ones:
  - `f25a7e0c` feat(spawn): spawned agents self-compact
  - `60a40e2` fix(spawn): pin core.database to backend DB
- `60a40e2` contains **both** halves of the fix (verified: `attach_compaction_hooks_to_all` in `isolated_runner.py:617`, `JARVIS_DB_PATH` in `isolated_spawner.py:219`).

### Test — passes, guards the right thing
`tests/test_services/test_spawned_compaction.py` correctly lives in the jarvis backend (it imports `services.context_compaction`, a backend module, so it can't live in the fast-agent repo). It pins the three properties the spawned path depends on: merge ordering (`["spawn","compact"]`), idempotency (`==1` then `==0`), and `on_context_overflow` survival. **2 passed** locally.

### Finding

🟡 **Stale PR body — the dependency note is already resolved.** The body still says *"Depends on fast-agent#13 (merge first); submodule pointer = `f25a7e0c`, re-point to the merge commit before merging here."* Reality:
- #13 is **already merged**.
- The pointer is **already** at `60a40e2` (one commit past `f25a7e0c` — it includes the DB-pin commit, which is good).

So the TODO is done; the body just reads as if it isn't. Harmless, but update it (or merge as-is) so a reviewer doesn't block waiting on a dependency that's already satisfied.

🟢 Minor/optional: the pointer is the branch-tip commit (`60a40e2`) rather than the merge commit (`2d56b7c9`). Both are on `main` with identical trees for these files, so it works either way — no action needed.

No blockers.

PR #120, opened 2026-07-01
feat(memory): dense-index fail-loud health + one-click repair; conflict-aware recall

omnigentx:
## Why
Two memory-subsystem gaps surfaced (and were reproduced) this session:

1. **Silent dead dense recall.** LadybugDB's HNSW vector index goes dead after an ungraceful process kill (agents/OOM/container stop) — rows stay but `QUERY_VECTOR_INDEX` returns nothing. This is a LadybugDB/Kùzu WAL bug **still unfixed through ladybug 0.18.0** (measured; see [kuzudb/kuzu#5184](https://github.com/kuzudb/kuzu/issues/5184)). `index-status` read "healthy" (count>0, store open) the whole time → recall silently degraded to keyword-only with **no signal**.
2. **Conflicting memories** ("works at X" vs "works at Y") surfaced together with no way for the agent to pick, and no UI to manage them.

### Memory-index resilience (core — verified end-to-end on a live instance)
- **Fail-loud health probe** — `index-status` now *self-queries* the HNSW index (with a stored vector, no embedding model needed), not just `count>0`. A present-but-dead index reports `dense.searchable=false` / `dense.broken=true`.
- **UI: red banner + one-click Restore** — Agent Memory panel shows a loud banner when dense recall is down, with a **Restore index** button → `POST /memory/repair` (DROP+CREATE the index **then** re-project from SQLite; re-upsert alone does NOT revive a dead index — measured). Bilingual EN/VI, spinner, clears on success.
- **Prevention: `CHECKPOINT` after every store write** — a kill then only ever lands on a consistent on-disk state (measured: checkpoint-then-kill reopens clean vs SIGSEGV without).
- **Bug fix:** re-upserting the *sole* memory (`DETACH DELETE` → empty → `CREATE`) killed the index and stayed dead (reproduced 12/12; this is what a single-memory re-projection hit). `upsert_memory` now rebuilds when the table is back to one row — the insert-side mirror of `delete_memory`'s delete-to-empty guard.

### Conflict-aware recall
- `retrieval_hook` stamps each recalled memory's **saved date** into the block the LLM reads → recency-tiebreak.
- Agent Memory panel: **"superseded" filter** + per-memory **version history** + **rollback**.
- `services/memory/conflict.py`: best-effort, background, embedding-gated + LLM-judged supersession of a same-slot contradiction on capture.

## Verification
- **82 unit/integration tests pass** (real `LadybugStore`, no mocks): index-health probe detects a dead index; re-upsert-sole regression; conflict resolver (real DB + real supersede); provider/indexing/retrieval-core/orchestrator/settings-route.
- **Live click-flow:** broke the index → red banner appeared → clicked **Restore index** → banner cleared + dense recall live again (`dense_hits=1`, `broken=false`).

## Known limitation / follow-up
- `conflict.py` is **best-effort**: it can race with the agent's own `memory_forget` (caught, logged, no data loss). A slot-level idempotency guard is a documented follow-up.
- No automated Playwright e2e for the banner yet (verified manually via the real UI).


backend/services/memory/conflict.py
```diff
@@ -0,0 +1,208 @@
...
+
+Return ONLY JSON, no prose, no code fence:
+{{"superseded": [<numbers of EXISTING that the NEW one replaces>]}}
+Empty list if none.
+"""
+
+
+def _shared_embed(texts: list[str]) -> list[list[float]]:
+    from services.indexing.embedding_provider import get_shared_embedding_provider
+    return get_shared_embedding_provider().embed_documents(texts)
```
omnigentx: **[HIGH] Wrong embedding model — silently disables conflict resolution on the default config.**

`_shared_embed` calls `get_shared_embedding_provider()` with **no args**, so it uses `DEFAULT_MODEL = "BAAI/bge-m3"` (`embedding_provider.py:20,229`). But the configured default embedding model is `"Qwen/Qwen3-Embedding-0.6B"` (`services/memory/settings.py:45`), and every other caller passes it explicitly — `orchestrator.py:53` and `memory_index_worker.py:92` both call `get_shared_embedding_provider(settings.embedding_model, settings.embedding_revision)`.

The shared provider is a **singleton keyed by `(model_name, revision)` that rebuilds `_SHARED` whenever the key changes** (`embedding_provider.py:233-236`). Consequences on the default config:

- **Model thrash:** every conflict-gate embed swaps the shared singleton `Qwen → bge-m3`, and the next retrieval/index op swaps it back — repeatedly reloading a large model.
- **Silent feature death:** if `bge-m3` isn't downloaded, `_shared_embed` raises → `_gate` swallows it (`return []`) → `resolve_conflicts` no-ops. Conflict resolution **never fires in production**, with no signal. The unit tests miss this because they inject `embed_fn`, bypassing `_shared_embed` entirely.

**Fix** — mirror the other callers:
```python
def _shared_embed(texts):
    from services.memory.settings import get_memory_settings
    from services.indexing.embedding_provider import get_shared_embedding_provider
    cfg = get_memory_settings()
    return get_shared_embedding_provider(
        cfg.embedding_model, cfg.embedding_revision).embed_documents(texts)
```
Also worth a test that exercises the real `_shared_embed` path (not the injected `embed_fn`) so a future model-key drift is caught.

backend/routes/memory.py
```diff
@@ -340,6 +340,33 @@ async def reindex() -> dict[str, Any]:
...
+    source of truth). Returns whether the index is searchable again so the UI can
+    clear the banner or keep failing loud."""
+    cfg = get_memory_settings()
+    rebuilt = False
+    err = None
+    try:
+        from services.indexing.ladybug_store import get_ladybug_store
+        store = get_ladybug_store(cfg.ladybug_path)
+        store.rebuild_vector_index()
+        rebuilt = store.probe_index_healthy()
```
omnigentx: **[LOW] `index_rebuilt` is probed before re-projection finishes.**

`rebuilt = store.probe_index_healthy()` runs right after `rebuild_vector_index()`, but `cs.rebuild(db, …)` (re-project from SQLite) runs *after* this and never re-checks health, so the returned `index_rebuilt` can be stale relative to the final state. Low impact — it only drives the toast; the banner re-probes via the frontend's `loadAll()` → `index-status`. Consider re-probing after `cs.rebuild` so the returned status reflects the end state.

omnigentx (commented):
GitNexus PR review — 2 findings (verified against the code, not the diff alone).

**Verified OK:** repair route/prefix, versions/rollback endpoints + payload shape, `current_version` in list, `cs.rebuild`/`supersede_memory` signatures, `probe_index_healthy` has no injection, regression tests are meaningful.

Requesting changes on **#1** only (silent-fallback that kills conflict-resolution on the default config). #2 is optional polish.

omnigentx:
Both findings addressed in be42332:

**#1 [HIGH] — gate embedder** — `_shared_embed` now passes `cfg.embedding_model`/`cfg.embedding_revision` to `get_shared_embedding_provider(...)` (matching `orchestrator.py:53` + `memory_index_worker.py:92`), so the gate embeds with the configured **Qwen3-Embedding-0.6B** — same space as the indexed memories — instead of the bge-m3 default. Added `test_shared_embed_uses_configured_model` to pin it.

**#2 [LOW] — repair probe order** — `/memory/repair` now runs `probe_index_healthy()` **after** both `rebuild_vector_index()` and `cs.rebuild()`, so the returned `index_rebuilt` reflects the final state.

21 tests pass (conflict + settings route), ruff clean.

PR #121, opened 2026-07-02
fix(memory): conflict gate reuses stored vectors + runs off the event loop

omnigentx:
## Root cause (verified live on prod)

`memory_remember` showed **"Unknown error"** in the UI. Root cause: the conflict-resolution gate (`conflict._gate`) **re-embedded the ENTIRE `(owner, scope, type)` memory slot on every save** — one sentence-transformers batch, padded to the longest fact → cost **O(N × max_len)**. On a 14-memory store this measured **~150s of pure CPU**, and because `schedule_resolve_conflicts` runs it as a task on the **main asyncio loop**, the synchronous embed **froze the whole event loop** → the live chat turn's SSE hung → UI rendered "Unknown error" (the memory was still saved).

### Evidence
- **Live**: a trivial async endpoint (`/api/health`) hung **71s**, unblocking the instant the embed finished (`tool_done duration=133s`), then back to `0.001s`.
- During the stall the **main thread (tid=1) + 3 torch threads were pinned at 100% CPU** — genuine compute, not I/O/lock.
- Diagnostic caller stack: `resolve_conflicts → _gate → _shared_embed → embed_documents(n=15, max_chars=1306, took=153.8s)`.
- Isolated: sync embed on a real loop → loop lag == embed time; `asyncio.to_thread` → **0.00s lag**.

## Fix
The module header already anticipated this: *"if a slot grows past this, switch the gate to the dense index (vector search)."*
- `_gate` production path (`embed_fn=None`) embeds **only the new fact** and scores candidates against their **already-stored** vectors via `LadybugStore.vector_search` (`_stored_scores`). Falls back to embedding if the dense index is unavailable. Injected-`embed_fn` tests keep the direct path.
- `resolve_conflicts` runs the CPU-bound gate via `asyncio.to_thread` → never blocks the loop.

Result: ~150s → ~0.3s, and the turn no longer stalls.

**LOW** — `resolve_conflicts` upstream is only `schedule_resolve_conflicts` (fire-and-forget); change is contained to `conflict.py`.

## Tests
- `test_gate_production_reuses_stored_vectors` — only the new fact is embedded; candidates scored from the index.
- `test_gate_production_falls_back_when_index_unavailable`.
- Existing scripted-`embed_fn` tests unchanged. **14 passed.**

Not yet verified end-to-end on prod (needs deploy) — see PR discussion.

### DIFF
--- a/backend/services/memory/conflict.py
+++ b/backend/services/memory/conflict.py
@@ -38,8 +38,14 @@ import asyncio
 import json
 import logging
 import re
+from collections import namedtuple
 
 logger = logging.getLogger("memory.conflict")
+
+# Plain, thread-safe carrier for the gate. resolve_conflicts converts the loaded
+# ORM rows to these BEFORE handing them to the to_thread worker so the worker never
+# touches a main-thread-owned Session (PR #121 review #2).
+_GateCand = namedtuple("_GateCand", ("id", "content", "owner_agent_name"))
 
 # Loose gate: only PURPOSE is to make the LLM call rare by skipping saves with no
 # semantically-close sibling. The LLM is the real judge, so erring low here is
@@ -83,6 +89,10 @@ def _shared_embed(texts: list[str]) -> list[list[float]]:
     # Qwen3-Embedding-0.6B — the cosine gate would then be meaningless and silently
     # disable conflict resolution on the default config (PR #120 review). Mirrors
     # orchestrator.py / memory_index_worker.py, and reuses the same warm singleton.
+    # MUST stay identical to the INDEXER's embedder (model + normalization): the
+    # stored-vector gate (_stored_scores) compares a fresh qv against index vectors,
+    # so a silent model/normalization divergence would mis-gate WITHOUT erroring
+    # (vector_search only catches a dimension mismatch) (PR #121 review #3).
     from services.indexing.embedding_provider import get_shared_embedding_provider
     from services.memory.settings import get_memory_settings
     cfg = get_memory_settings()
@@ -178,6 +188,20 @@ def _gate(new_content: str, cands: list, embed_fn) -> list:
     scores = _stored_scores(cands, qv)
     if scores is None:                    # dense index unavailable → re-embed fallback
         return _gate_by_embed(new_content, cands, _shared_embed)
+    # vector_search ranks across the owner's WHOLE store (owner+status only, not the
+    # slot), so once an owner grows past the fetch window a genuinely-similar same-slot
+    # candidate can fall outside `hits` — scoring it 0.0 would silently drop it and miss
+    # the conflict (PR #121 review #1). Embed ONLY those few directly. The common case
+    # (all cands are near neighbours, or a small store) embeds nothing here.
+    missing = [c for c in cands if c.id not in scores]
+    if missing:
+        logger.info("[MEMORY] conflict gate: %d/%d candidate(s) outside the dense "
+                    "window — scoring by embedding", len(missing), len(cands))
+        try:
+            for c, v in zip(missing, _shared_embed([c.content for c in missing])):
+                scores[c.id] = _cosine(qv, v)
+        except Exception as exc:  # noqa: BLE001 — best-effort; keep the dense scores
+            logger.warning("[MEMORY] conflict gate tail-embed failed: %s", exc)
     return _top([(c, scores.get(c.id, 0.0)) for c in cands])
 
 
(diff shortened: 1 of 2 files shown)


omnigentx:
## PR Review — conflict gate reuses stored vectors + runs off the event loop

**Risk: LOW–MEDIUM** · +121/−10 in `conflict.py` + tests. Root cause is correctly identified and the two-part fix is sound. Approving, with one scaling concern worth addressing.

### Verified by tracing the code (not just the diff)
- **`scores.get(c.id)` actually matches.** The dense index keys Memory nodes by `record_id`, and the indexer sets `record_id = rec.id` (`memory_index_worker.py:243`); `vector_search` returns `node.id`. So `MemoryRecord.id == VectorHit.record_id` — the production path won't silently score everything 0.0 and return `[]`. ✅
- **Off-loop is genuinely safe.** `LadybugStore.vector_search` runs under `self._lock` (RLock), so overlapping fire-and-forget `resolve_conflicts` tasks hitting the store from `to_thread` workers are serialized. ✅
- **Distance→sim is right.** Index is built `metric := 'cosine'`; `similarity = 1 - distance` matches. ✅
- **Fallbacks intact.** Dense outage → `_stored_scores` returns `None` → `_gate_by_embed`; injected `embed_fn` keeps the direct path so scripted tests are unchanged. ✅

### Findings

1. **[medium] Global-owner ranking can drop a same-slot candidate → missed dedup as the store grows.**
   `cands` is the same `(owner, subject_scope, memory_type)` slot (≤40). But `_stored_scores` calls `vector_search(owner=…, limit=40)`, which ranks across **all** of the owner's memories (every scope/type) — the `WHERE` filters only `owner` + `status`, not the slot. It returns the 40 globally-nearest. Once an owner accumulates **>40** memories and other-slot facts crowd the top-40, a genuinely-similar same-slot candidate ranked >40 gets `scores.get(c.id, 0.0) = 0.0`, is cut by the 0.45 floor, and the conflict is **never detected → the duplicate persists silently** (no error). The old path scored all ≤40 slot candidates directly, so this is a behavioral divergence that worsens exactly along the scale axis the module header cared about. For a small store (the 14-memory repro) it's invisible; for the long-lived primary `user` scope it will bite.
   *Suggestions:* filter the dense query by `memory_type` (and scope if the node carries it) to match the slot, or — for any candidate absent from `hits` — score it against its stored vector directly instead of assuming 0.0. At minimum, `log` when `len(hits) < len(cands)` so the silent drop is observable.

2. **[low] Session-bound ORM rows crossed into a worker thread.**
   `await asyncio.to_thread(_gate, rec.content, cands, embed_fn)` hands live `MemoryRecord` objects to another thread, which reads `c.content`, `c.id`, `c.owner_agent_name`. Safe **today** — those columns are eagerly loaded by the SELECT and nothing commits/expires them before the gate. But it's fragile: if a later edit adds a commit (default `expire_on_commit=True`) or a lazy attribute before this line, the first off-thread access triggers a lazy reload on a session owned by the main thread → non-thread-safe `Session` used concurrently. Cheap hardening: pass plain tuples/dataclasses `(id, content, owner)` into the thread rather than the ORM rows.

3. **[info] Embedding-space coupling.** The stored-vector path compares a freshly-`_shared_embed`'d `qv` against index vectors; correctness assumes `_shared_embed` and the indexing embedder stay identical (model + normalization). They are today, and `vector_search` raises on `EMBED_DIM` mismatch (catches dimension drift), but the 0.45 gate is delicate given the known bge-m3 name-similarity range (~0.32–0.40) — a silent model divergence would mis-gate without erroring. Worth a one-line "must match the indexer" comment near `_shared_embed`.

4. **[info] CI doesn't exercise the real ladybug contract.** The new tests use a `FakeStore`, so `record_id`/`distance` correctness is mocked — I verified it by hand above, but it's unguarded against future index-schema changes. Given this gate caused a 150s prod loop-stall, a prod smoke (save into a >5-memory slot; confirm turn latency is back to ~0.3s **and** a real duplicate still supersedes) before/after deploy is worth it.

### Recommendation
**APPROVE.** The loop-stall root cause is fixed correctly (whole-slot re-embed removed; gate moved off the event loop) and I confirmed the id/lock/metric contracts the fix relies on. Finding #1 is the one I'd either fix (slot-filter the dense query) or consciously accept and log before it scales.

_Reviewed with the GitNexus PR-review workflow; contracts verified against `ladybug_store.py`, `memory_index_worker.py`._

omnigentx:
Addressed the review in `f51c75e`:

- **#1 (medium — silent same-slot drop):** fixed. `vector_search` ranks across the owner's whole store, so a same-slot candidate outside the fetch window would have scored `0.0` and been dropped. `_gate` now embeds **only** the candidates absent from the dense window (and `log`s how many), so no same-slot candidate is silently dropped as the store grows; the common case still embeds nothing beyond the new fact. New test: `test_gate_embeds_candidates_missing_from_dense_window`.
- **#2 (low — ORM row across thread):** fixed. The gate now crosses the `to_thread` boundary with plain `_GateCand` tuples `(id, content, owner_agent_name)` built on the main thread, so the worker never risks a lazy reload on the main-thread `Session`.
- **#3 (info — embedding-space coupling):** added a comment on `_shared_embed` that it must stay identical to the indexer's embedder (model + normalization), since the stored-vector gate compares against index vectors.
- **#4 (info — real ladybug contract untested):** acknowledged; the new tests still use `FakeStore`. Will run a prod smoke (save into a >5-memory slot → turn latency back to ~0.3s **and** a real duplicate still supersedes) as part of deploy verification.

15 tests pass.

PR #125, opened 2026-07-03
perf(tts): SQLite SSoT for pre-gen chapter status; stop per-tick full disk scan

omnigentx:
## Problem

The TTS pre-gen scheduler re-derived *"which chapters need audio"* from disk on **every idle tick (~10s)**: read every chapter's text → `clean_text_for_tts` + md5 → `os.path.exists` the cache file — for the whole story tree, **inline on the event loop**.

Measured on production: **~1.0s of blocking CPU/IO every ~11s** (one story, 561 chapters; 5.2MB re-read + 561 hashes each pass). It froze the event loop in ~1s bursts (visible as `/api/health` stalling ~1s every 11s) and burned ~8600 file-ops/min **forever, even when everything was already generated**.

Root reason: no durable per-chapter status, and the content hash (the expensive part) was recomputed for immutable files on every pass.

## Redesign — SQLite as the single source of truth

- **New `story_chapters` table** (PK `story_id`+`chapter_file`): `chapter_num`, `content_hash`, text `(mtime, size)` fingerprint, `status` (`pending`|`ready`).
- **`services/story_audio_index.py`:**
  - `reconcile(db, sig)` — sync table↔disk, re-reading/re-hashing **only** files whose `(mtime, size)` changed; a position shift just updates `chapter_num`; deleted files drop out. Steady state = one `stat` per chapter, no read/hash.
  - `next_task` / `preview_queue` — priority (P0→P3 via `story_progress`) as **indexed queries**, no disk, no hashing.
  - `status_counts` — done/total as a `COUNT` → instant status endpoint.
  - `mark_ready` / `mark_pending_by_hash` — status from generate / evict.
- **`tts_pregen_job.py`:** `get_next_task` runs the fingerprint + any reconcile + the disk-quota sweep on a worker thread (`asyncio.to_thread`) so nothing blocks the loop; picking the task is one query. `execute_task` flips status→`ready`; eviction flips evicted hashes→`pending`; `rescan()` (called by the stories API on add/remove) forces a reconcile — the event-driven trigger. Dead P0–P3 finder/collector scan methods removed.

## Effect

- Steady-state per tick: **~1s → ~ms**, and off the event loop.
- Status check: a DB count, not a disk scan.
- Cost scales with **changed** files, not total chapters.

**LOW** — `TTSPreGenJob` upstream is only `server.py` lifespan; the stories route / scheduler use the preserved public methods (`rescan`/`get_next_task`/`preview_queue`/`get_status`/`execute_task`).

## Tests

- `test_story_audio_index.py`: reconcile build / skip-unchanged (no re-hash) / add+remove, P0/P2/P3 priority, mark_ready + evict→pending, preview order — **8 passed**.
- Existing `test_tts_pregen_integration.py` still green.
- New table auto-creates via `Base.metadata.create_all`; first boot runs one reconcile (backfills status from existing mp3s on disk — no bulk re-hash of already-generated chapters beyond the one-time pass).

### DIFF
--- a/backend/core/database.py
+++ b/backend/core/database.py
@@ -161,6 +161,28 @@ class AudioCacheEntry(Base):
     last_accessed_at = Column(Float, nullable=True)  # for LRU eviction
 
 
+class StoryChapter(Base):
+    """SSoT for per-chapter TTS pre-generation status.
+
+    Maps a chapter ``(story_id, chapter_file)`` to the content hash of its cleaned
+    text plus a file fingerprint, and tracks whether its audio is generated. The
+    reconcile pass (services.story_audio_index) keeps this in sync with
+    ``data/stories`` + the audio cache on disk — re-reading/re-hashing ONLY files
+    whose ``(mtime, size)`` changed. The pre-gen queue and status reads then hit
+    ONLY this table (indexed), replacing the old per-tick full disk scan that
+    re-read + re-hashed every chapter (~1s of blocking work every 10s)."""
+    __tablename__ = "story_chapters"
+
+    story_id = Column(String(255), primary_key=True)      # = folder under data/stories/
+    chapter_file = Column(String(255), primary_key=True)  # .txt filename
+    chapter_num = Column(Integer, default=0)              # sorted position within story
+    content_hash = Column(String(64), nullable=True, index=True)  # md5(clean_text) → cache key
+    text_mtime = Column(Float, default=0.0)               # fingerprint: source .txt mtime
+    text_size = Column(Integer, default=0)                # fingerprint: source .txt size
+    status = Column(String(20), default="pending", index=True)  # pending | ready | failed
+    updated_at = Column(Float, default=lambda: datetime.now().timestamp())
+
+
 class PendingAction(Base):
     """Cross-process action queue — replaces pending_read.json.
     MCP subprocess writes; FastAPI routes read and delete."""
--- a/backend/routes/stories.py
+++ b/backend/routes/stories.py
@@ -245,11 +245,12 @@ async def delete_story_endpoint(story_id: str, _=Depends(verify_api_key)):
         logger.error(f"Delete failed: {e}", exc_info=True)
         return JSONResponse(status_code=500, content={"error": str(e)})
     finally:
-        # Trigger TTS pregen rescan
+        # Trigger TTS pregen rescan — OFF the event loop: rescan() stats the whole
+        # story tree + reconciles, which shouldn't block the loop on a user action.
         if _state.bg_scheduler:
             for job in _state.bg_scheduler._jobs:
                 if hasattr(job, 'rescan'):
-                    job.rescan()
+                    await asyncio.to_thread(job.rescan)
 
 
 @router.get("/crawl/status")
--- a/backend/services/story_audio_index.py
+++ b/backend/services/story_audio_index.py
@@ -20,8 +20,10 @@ This module makes SQLite the single source of truth for per-chapter status:
 
 Readiness is authoritative in this table (``status``): reconcile seeds it from the
 mp3's existence for new/changed chapters; generation flips it to ``ready``; eviction
-flips it back to ``pending``. The serve path still checks the file directly, so an
-out-of-band mp3 deletion is caught there; a ``force`` reconcile re-syncs drift.
+flips it back to ``pending`` (mark_pending_by_hash). An out-of-band mp3 deletion
+(the ``.txt`` unchanged, so the fingerprint can't see it) is re-synced by the force
+reconcile (``verify_ready``, run by ``rescan``); the serve path also regenerates
+on demand when it finds the file missing.
 """
 from __future__ import annotations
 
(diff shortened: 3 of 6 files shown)


omnigentx:
## PR Review — SQLite SSoT for TTS pre-gen chapter status

**Risk: LOW–MEDIUM** · +573/−306 across 4 files. Strong redesign: the hot per-tick full-tree read+hash is genuinely gone, and the content-hash invariant that the whole thing rests on is airtight. Two loop/robustness gaps worth fixing before merge; the rest is polish. Ran the suite — **10 passed** (8 new `test_story_audio_index` + 2 `test_tts_pregen_integration`).

### Verified by tracing, not just reading
- **The hash SSoT is consistent end-to-end.** `reconcile` (`_hash_and_status`), `mark_ready` (`get_content_hash(cleaned_text)`), the cache filename (`get_audio_cache_path`), and eviction (`file[:-4]` → `mark_pending_by_hash`) all key on `md5(clean_text_for_tts(text))`. Both read paths — `story_audio_index._read_chapter` and the job's `_read_chapter_text` — `.strip()` before cleaning, so the two sources can't hash-diverge. This is the load-bearing invariant; it holds. ✅
- **Hot path is off-loop.** `get_next_task` runs `_maybe_reconcile` + `_ensure_disk_quota` via `asyncio.to_thread`; task selection is one indexed query. Steady state = one `os.stat` per chapter (the `signature()` fingerprint), no read/hash — exactly the claimed ~1s→~ms. ✅
- **Change detection is sound.** `signature()` compares `(story, ch, mtime, size)`; reconcile re-hashes only rows whose `(mtime,size)` moved, a position shift just rewrites `chapter_num`, deletes drop out. ✅

### Findings

1. **[medium] `rescan()` runs the full-tree fingerprint + reconcile *inline on the event loop*.**
   `stories.py:252` calls `job.rescan()` in the `finally` of `async def delete_story_endpoint`. `rescan()` → `signature()` does one `os.stat` for **every chapter in the whole tree** (561 on the prod story) plus a DB reconcile+commit, synchronously on the loop. This is the exact class of work the PR moves off-loop everywhere else. It's far milder than the old per-tick scan — it's a rare user action and a delete re-hashes nothing — but it's an inconsistent, tree-sized stat sweep on the loop. The route is already `async`, so: `await asyncio.to_thread(job.rescan)` (or have `rescan` offload internally).

2. **[low–medium] Concurrent reconcile is unguarded and can raise / interleave.** `rescan()` (loop thread) and `_maybe_reconcile()` (worker thread from `get_next_task`) can run at once, each with its own session committing to `story_chapters`. `rescan` swallows exceptions, but `get_next_task` does **not** wrap `await asyncio.to_thread(self._maybe_reconcile)` — a transient SQLite "database is locked" there propagates into the scheduler instead of just skipping a cycle, and two overlapping reconciles can interleave row writes. Guard `_maybe_reconcile` like `rescan` does, and/or add a single reconcile lock so only one runs at a time.

3. **[low] Module docstring overstates recovery from an out-of-band mp3 deletion.** The header says *"a `force` reconcile re-syncs drift"* for a deleted mp3 — but `reconcile` only re-derives `status` when the **.txt** `(mtime,size)` changes. Deleting the mp3 doesn't touch the `.txt`, so `status` stays `ready` forever and the chapter is never re-queued; only the serve-path file check saves it (on-demand regen). Internal eviction is fine (it calls `mark_pending_by_hash`), so the real gap is narrow (manual/out-of-band deletion), but the comment claims a recovery that doesn't happen. Fix the comment, or add an mp3-existence check to the `force` path.

4. **[info] The new threading/concurrency paths are untested.** `test_story_audio_index` is excellent (real SQLite, real reconcile), but it exercises the *index module* only — the job-level `to_thread` offload, the on-loop `rescan`, and the rescan↔reconcile race (#1/#2) have no coverage; the integration test predates them. A test that fires `get_next_task` concurrently with a `rescan` would lock in #2.

### Nits (non-blocking)
- `self._stats["done_chapters"] += 1` in `execute_task` is transient bookkeeping — `get_status`/`status_counts` recompute from the DB (the SSoT) each call, so the increment can briefly drift but is always corrected. Harmless; could drop it.
- Priority-label mismatch: `preview_queue` labels the first pending chapter **P0** even when it isn't literally `last_num+1` (e.g. n+1 already ready), while `next_task` reserves P0 for `chapter_num == last_num+1`. Same chapter gets generated either way — only the SSE snapshot's priority number differs.

### Recommendation
**APPROVE with nits.** The core is a clean, correct SSoT and the hot path is genuinely off-loop. Please address #1 (`to_thread` the `rescan`) and #2 (guard/serialize concurrent reconcile) — both keep the loop clean and the scheduler resilient, in the spirit of the PR itself. #3/#4 are quick follow-ups.

_Reviewed with the GitNexus PR-review workflow; contracts verified against `story_audio_index.py`, `tts_pregen_job.py`, `routes/stories.py`, `helpers/audio_cache.py`. Tests run locally: 10 passed._

omnigentx:
Added `test_tts_pregen_e2e.py` — a happy-path e2e that drives the **real** `TTSPreGenJob` end to end (construction/reconcile → `get_next_task` with the real `asyncio.to_thread` reconcile + indexed SSoT query → `execute_task` real already-on-disk fast path → status write → `get_status` → queue advance, plus new-story pickup on the next tick), against real in-memory SQLite + a temp story tree/audio cache. Only the edge_tts network is avoided (via the already-done short-circuit); the queue/status subsystem under change runs for real. 12 tests pass (2 e2e + 8 SSoT + 2 existing integration).

omnigentx:
Addressed the review in `e61b769`:

- **#1 (medium — on-loop rescan):** the stories delete route now `await asyncio.to_thread(job.rescan)`, so the whole-tree stat sweep + reconcile runs off the event loop.
- **#2 (low–med — unguarded concurrent reconcile):** added a `threading.Lock` that serialises `rescan()` and `_maybe_reconcile()`; `_maybe_reconcile` re-checks the fingerprint under the lock and now swallows a transient error (a `database is locked` just skips the cycle instead of reaching the scheduler).
- **#3 (low — drift claim):** `reconcile(verify_ready=True)` (used by `rescan`) re-checks each `ready` chapter's mp3 and requeues an out-of-band deletion; the module docstring is corrected to state exactly this.
- **nit:** `preview_queue` now labels `P0` only when `chapter_num == last_num+1` (matches `next_task`).

New tests: `verify_ready` requeues a deleted mp3; an e2e that races `get_next_task` against `rescan` (concurrent) and asserts it neither raises nor corrupts the table. **14 passed** (2 e2e-flow + 1 e2e-concurrency + 9 SSoT + 2 existing integration).

PR #127, opened 2026-07-03
fix(iot): store Roborock login session in the DB (encrypted), not a git-cleaned file

omnigentx:
## Problem

The Roborock login session (a **secret**) was persisted as a plaintext JSON file at `config/credentials/roborock_session.json`. That directory is bind-mounted from the CI git checkout, and `deploy.yml`'s `actions/checkout` runs with **`clean: true`** → `git clean -ffdx`, which **wipes the (gitignored) session on every deploy**. So after each redeploy the user had to redo Roborock **2FA**. It was also stored in plaintext, unlike every other credential in the app.

Confirmed on prod: `config/credentials/` was empty (mtime = last deploy), `roborock_session.json` absent.

## Fix

Persist the session in the **DB via `config_service` with `is_secret=True`** (encrypted at rest), category `service.roborock` — the same pattern as Google OAuth (`oauth.google`), Gmail, GitHub PAT (`service.github`) and gateway bot tokens. `system_config` lives in `jarvis.db` on the **`data/` Docker volume**, which `git clean` never touches → the session **survives deploys**.

Verified the MCP subprocess can use this: `config_service` is already called from other tool subprocesses (`skill_server`, `cron_server`, `mcp_admin_server`), and `JARVIS_MASTER_KEY` is present in the container (encrypt/decrypt works).

## Audit (requested: is anything else stored in a git-cleaned path?)

**No — Roborock was the only offender.** Everything else is already DB-backed via `config_service`:

| Credential | Storage | Wiped by deploy? |
|---|---|---|
| Roborock | ~~file `config/credentials/`~~ → **DB `service.roborock` (encrypted)** | fixed |
| Google (Gmail/Calendar) | DB `oauth.google` (encrypted) | no |
| Gmail | DB (encrypted) | no |
| GitHub PAT | DB `service.github`; `hosts.yml` is a regenerated surface | no |
| Telegram / Zalo bot | DB (`is_secret=True`) | no |
| `fastagent.secrets.yaml` | separate host bind mount (not the checkout) | no |

**LOW** — `save_session`/`load_session` are called only within `iot_server.py`; all 13 downstream callers (get_session → list_devices/get_target_device → the robot command tools) are transparent (storage backend changed file→DB, same signatures).

## Tests

`test_iot_roborock_session.py`: session persists under `service.roborock` with `is_secret=True` (encrypted), round-trips via the DB, `load` returns `None` when absent. **3 passed.**

## Note

No migration: the prod session file is already gone (git-cleaned), so the user re-does 2FA **once** after this ships, then it lives in the encrypted DB permanently across all future deploys.

### DIFF
--- a/backend/fastagent.config.yaml
+++ b/backend/fastagent.config.yaml
@@ -142,6 +142,12 @@ mcp:
          PYTHONUNBUFFERED: "1"
          ROBOROCK_USERNAME: "${ROBOROCK_USERNAME}"
          ROBOROCK_PASSWORD: "${ROBOROCK_PASSWORD}"
+         # The Roborock session persists to the DB via config_service (encrypted).
+         # Pin BOTH bindings this depends on so a later `cwd:`/chdir can't silently
+         # point config_service at a different jarvis.db, and the master key is
+         # available to decrypt (mirrors meeting_room + the google/other blocks).
+         JARVIS_DB_PATH: "data/jarvis.db"
+         JARVIS_MASTER_KEY: "${JARVIS_MASTER_KEY}"
     scrapling-server:
       command: "scrapling"
       args: ["mcp"]
--- a/backend/tests/test_services/test_iot_roborock_session.py
+++ b/backend/tests/test_services/test_iot_roborock_session.py
@@ -64,3 +64,32 @@ def test_session_round_trips_through_db(monkeypatch):
 def test_load_returns_none_when_nothing_stored(monkeypatch):
     _use_fake_config(monkeypatch)
     assert asyncio.run(iot_server.load_session()) is None
+
+
+def test_session_encrypts_at_rest_via_real_config_service(monkeypatch, tmp_path):
+    """End-to-end crypto: with a real ConfigService + real Fernet, is_secret=True
+    stores CIPHERTEXT in system_config and load_session decrypts it back."""
+    from core import secrets_crypto
+    from core.database import Base, SystemConfig
+    from sqlalchemy import create_engine
+    from sqlalchemy.orm import sessionmaker
+    from services.config_service import ConfigService
+
+    monkeypatch.setenv("JARVIS_MASTER_KEY", "iot-tests-master-key-please-rotate")
+    secrets_crypto.reload_master_key()
+    engine = create_engine(f"sqlite:///{tmp_path}/cfg.db", future=True)
+    Base.metadata.create_all(engine)
+    Session = sessionmaker(bind=engine, future=True, expire_on_commit=False)
+    monkeypatch.setattr("services.config_service.config_service",
+                        ConfigService(db_factory=Session))
+    monkeypatch.setattr(iot_server, "UserData",
+                        type("UD", (), {"from_dict": staticmethod(lambda d: _FakeUserData(d))}))
+
+    ud = _FakeUserData({"token": "s3cr3t-value", "rriot": {"u": "x"}})
+    asyncio.run(iot_server.save_session(ud))
+
+    row = Session().query(SystemConfig).filter_by(
+        category="service.roborock", key="session").one()
+    assert row.is_secret is True
+    assert "s3cr3t-value" not in row.value                       # ciphertext at rest
+    assert asyncio.run(iot_server.load_session()).as_dict() == ud.as_dict()  # decrypts back
--- a/backend/tests/test_services/test_iot_roborock_session.py
+++ b/backend/tests/test_services/test_iot_roborock_session.py
@@ -1,0 +1,66 @@
+"""Roborock login session persistence (tools.iot_server).
+
+The session is a SECRET — it must persist to the DB via config_service with
+is_secret=True (encrypted at rest, survives deploys), like every other credential
+in the app. It used to be a plaintext JSON file under config/credentials/, which
+the deploy's `git clean` wiped every time → repeated 2FA. These lock in the DB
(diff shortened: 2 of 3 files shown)


omnigentx:
## PR Review — store Roborock session in the DB (encrypted), not a git-cleaned file

**Risk: LOW.** Correct root-cause fix and the right pattern — a secret belongs in the encrypted DB (`is_secret=True` → Fernet at rest), matching Google OAuth / GitHub PAT / bot tokens. It achieves the deploy-survival goal. Ran the tests — **3 passed**. Two hardening findings before merge, both about *not relying on env inheritance* for the very properties this fix depends on.

### Verified by tracing
- **Contract matches.** `config_service.set(cat, key, value, *, is_secret=False, …)` and `get(cat, key, *, default=None)` line up with the call sites; `is_secret=True` routes through `_encode → secrets_crypto.encrypt` (Fernet). `get()` returns `default` (→ `None`) when the row is absent, so `load_session`'s `if not raw: return None` is right. ✅
- **DB binding lands on the volume — today.** `iot-control` sets no `cwd`, so the subprocess inherits the server CWD (`/app` in Docker) and `core.database` resolves the relative `data/jarvis.db` to the **volume** DB — the same one the backend reads, which `git clean` never touches. So the session really does survive deploys. ✅
- **Blast radius is contained.** `save_session`/`load_session` are internal to `iot_server`; signatures unchanged (file→DB is transparent to the 13 downstream callers). The audit table (only Roborock was file-backed) is a good addition — I did not independently re-verify every row.

### Findings

1. **[low–medium] The DB binding is implicit (CWD inheritance), unlike every sibling subprocess — and this fix's whole value depends on it.** `iot-control` in `fastagent.config.yaml` sets neither `cwd` nor `JARVIS_DB_PATH`; it works *only* because it inherits CWD `/app` and resolves `data/jarvis.db` relatively. `meeting_room` (line 177) and the PR #13 spawned-agent fix both **explicitly pin `JARVIS_DB_PATH`** precisely to avoid this CWD-drift class of bug. If anyone later adds a `cwd:` to this block (or the launcher `chdir`s, like `isolated_runner` does), `config_service` silently writes to a *different* `data/jarvis.db` → the session neither persists nor is found, reintroducing the exact 2FA-every-deploy symptom this PR fixes, but harder to diagnose. **Add `JARVIS_DB_PATH: "data/jarvis.db"` to the `iot-control` env block** to make the binding explicit and match the established pattern.

2. **[low] `save_session` silently swallows a persist failure — and the new failure modes are more likely than a file write.** `encrypt()` **raises `MissingMasterKeyError`** when `JARVIS_MASTER_KEY` is absent (`core/secrets_crypto.py:80`), and `JARVIS_MASTER_KEY`, like the DB path, is only *inherited* by the iot subprocess, not set on the block. `save_session`'s `except Exception: print(...)` catches that (and any DB write error), so login proceeds as if saved while nothing persisted → the user silently re-does 2FA next process/deploy, with only a stdout `print` as evidence. The old file version swallowed too (not a regression), but a missing master key / DB error is a realer, quieter failure than `open().write()`. Recommend: `logger.error(..., exc_info=True)` (or re-raise) on save failure so a broken persist is visible, and set `JARVIS_MASTER_KEY` explicitly on the `iot-control` block alongside #1. (`load_session`'s swallow is fine — absent/undecryptable → `None` → re-login is correct degradation.)

3. **[info] Tests lock the call contract, not the real crypto or the subprocess DB binding.** The suite fakes `config_service` and `UserData`, so it verifies category/key/`is_secret`/JSON round-trip — good — but not (a) that a session truly encrypts+decrypts via real `secrets_crypto` against a `system_config` row, nor (b) that the iot subprocess binds to the volume DB. (a) is a cheap add: real `config_service` + temp DB + a real `JARVIS_MASTER_KEY`, round-trip `set(is_secret=True)`/`get`. (b) is config, covered by the explicit pin in #1. The PR's "verified in the container" is manual, not CI — worth stating.

### Recommendation
**APPROVE with nits.** The approach is correct and the goal is met. Please make the two bindings this fix silently depends on — `JARVIS_DB_PATH` and `JARVIS_MASTER_KEY` — **explicit on the `iot-control` block**, and stop swallowing a save failure silently. All three are quick and keep the fix from regressing invisibly.

_Reviewed with the GitNexus PR-review workflow; verified against `config_service.py`, `core/secrets_crypto.py`, `fastagent.config.yaml`. Tests run locally: 3 passed._

omnigentx:
Addressed all three in `65a4038`:

- **#1 (implicit DB binding):** pinned `JARVIS_DB_PATH: "data/jarvis.db"` on the `iot-control` MCP env block (mirrors `meeting_room`), so a later `cwd:`/chdir can't silently bind `config_service` to a different `jarvis.db` and reintroduce 2FA-every-deploy.
- **#2 (silent swallow + inherited master key):** also pinned `JARVIS_MASTER_KEY: "${JARVIS_MASTER_KEY}"` on the block; and `save_session` now `logger.error(..., exc_info=True)` on a failed persist instead of a silent `print` — a missing master key / DB error is now visible (not re-raised: the current session still works — degrade, don't break).
- **#3 (tests didn't exercise real crypto):** added an end-to-end test — real `ConfigService` + real Fernet + temp DB: `is_secret=True` stores **ciphertext** in `system_config` and `load_session` decrypts it back.

4 tests pass. (Subprocess→volume DB binding is now config, made explicit by #1.)

PR #128, opened 2026-07-07
feat(gateways): Telegram "/" command menu + immediate enable-toggle persistence

omnigentx:
## Why
Two gateway UX gaps surfaced while using the Telegram/Zalo bots:

1. **No slash-command discovery on Telegram.** The gateway has a command registry (`/new`, `/agent`, `/whoami`, `/help`) but never published it to Telegram, so typing `/` showed **no autocomplete** — users couldn't discover the commands. This is the exact affordance OpenClaw ships (it calls `setMyCommands`).
2. **The Enable toggle didn't persist on flip.** Toggling a gateway off then reloading reverted to the saved value and the gateway never actually stopped.

### 1. Telegram `/` command menu (`setMyCommands`)
- On startup (right after `getMe`) a registering gateway runs `deleteMyCommands` → `setMyCommands(commands.menu_commands())`. `_COMMANDS` stays the single source of truth — add a command there and it shows in the menu + `/help` for free.
- New flag `BotApiGateway.registers_commands` (default `False`); `TelegramGateway = True`.
- **Zalo stays `False`** — verified against **bot.zaloplatforms.com/docs**: the Zalo Bot API exposes only `getMe`/`getUpdates`/`setWebhook`/`sendMessage`/`sendPhoto`/`sendSticker`/`sendVoice`/`sendChatAction`, **no `setMyCommands`**. Registering there would just error every startup.
- Best-effort: a registration failure logs and never blocks the poll loop.

### 2. Enable toggle persists immediately
- The switch only mutated local state and required a separate **Save changes** click, so a flip-off then reload reverted (confirmed: `zalo_enabled` stayed `'true'` with no history entry, while the save path itself worked — `telegram_enabled=false` had saved fine).
- The toggle now saves immediately via `/settings/bulk` and **reverts on failure** so the UI never shows a state it didn't persist. Token / allow-list / agent still batch through **Save changes** (a switch applies now; form fields batch — the expected pattern).

## Verification
- **44 gateway tests pass** — added `test_telegram_registers_slash_command_menu` (asserts `deleteMyCommands` → `setMyCommands` with the canonical payload) and `test_zalo_does_not_register_commands`.
- **Toggle persistence verified end-to-end on a live instance**: `POST /settings/bulk zalo_enabled=false` → DB flipped to `false` **and** the Zalo gateway stopped (`running=False, connected=False`).
- Grounded, not guessed: read OpenClaw's `bot-native-command-menu.ts` (`setMyCommands` + `BOT_COMMANDS_TOO_MUCH` handling) and the official Zalo Bot API method list.

## Note
To see the menu on a live bot: enable Telegram (the gateway auto-registers on startup) then type `/` in the bot chat.

### DIFF
--- a/frontend/src/views/settings/SettingsGateways.vue
+++ b/frontend/src/views/settings/SettingsGateways.vue
@@ -165,7 +165,9 @@ function pollStatusAfterSave() {
 // never shows a state we didn't persist.
 async function toggleEnabled(id) {
   const r = rows[id]
-  r.error = ''
+  if (r.saving) return          // guard: a persist is already in flight (the switch
+  r.error = ''                  // is :disabled too, but belt-and-suspenders vs races)
+  r.saving = true
   try {
     await apiFetch('/api/settings/bulk', {
       method: 'POST',
@@ -177,6 +179,8 @@ async function toggleEnabled(id) {
   } catch (err) {
     r.enabled = !r.enabled  // failed to persist → don't lie about the state
     r.error = err?.body?.detail || err?.message || String(err)
+  } finally {
+    r.saving = false
   }
 }
 
@@ -267,7 +271,7 @@ onMounted(reload)
       <div class="card-body">
         <!-- Enable -->
         <label class="switch-row">
-          <input type="checkbox" v-model="r.enabled" @change="toggleEnabled(id)" />
+          <input type="checkbox" v-model="r.enabled" :disabled="r.saving" @change="toggleEnabled(id)" />
           <span class="switch-track"><span class="switch-thumb" /></span>
           <span class="switch-label">{{ t('settings.gateways.enable', { name: metaFor(id).label }) }}</span>
         </label>
--- a/frontend/src/views/settings/SettingsGateways.vue
+++ b/frontend/src/views/settings/SettingsGateways.vue
@@ -158,6 +158,28 @@ function pollStatusAfterSave() {
   setTimeout(tick, 800)
 }
 
+// The enable switch persists IMMEDIATELY on flip — a toggle is expected to APPLY,
+// not sit as a pending edit until "Save changes" (which is for token/allow-list/
+// agent). Without this, flipping off then reloading reverted to the saved value
+// and the gateway never actually stopped. Revert the toggle on failure so the UI
+// never shows a state we didn't persist.
+async function toggleEnabled(id) {
+  const r = rows[id]
+  r.error = ''
+  try {
+    await apiFetch('/api/settings/bulk', {
+      method: 'POST',
+      body: JSON.stringify({ items: [
+        { category: 'gateways', key: `${id}_enabled`, value: r.enabled ? 'true' : 'false', is_secret: false },
+      ] }),
+    })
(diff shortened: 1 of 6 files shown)


frontend/src/views/settings/SettingsGateways.vue
```diff
@@ -158,6 +158,28 @@ function pollStatusAfterSave() {
   setTimeout(tick, 800)
 }
 
+// The enable switch persists IMMEDIATELY on flip — a toggle is expected to APPLY,
+// not sit as a pending edit until "Save changes" (which is for token/allow-list/
+// agent). Without this, flipping off then reloading reverted to the saved value
+// and the gateway never actually stopped. Revert the toggle on failure so the UI
+// never shows a state we didn't persist.
+async function toggleEnabled(id) {
```
omnigentx, line 166: **[NIT — non-blocking] No in-flight guard on the toggle.**

`save(id)` sets `r.saving` to block double-submit, but `toggleEnabled` doesn't. Flipping the switch rapidly can fire overlapping `/settings/bulk` writes + two `pollStatusAfterSave()` runs. The final state stays correct (revert only on failure), so this is just redundant requests, not a correctness bug. If you want it tight: disable the checkbox while a persist is in flight (`:disabled` on an `r.saving`-style flag) so a flip can't overlap.

Also worth a small frontend unit test for the revert-on-failure path — the e2e was verified live but there's no automated coverage of it.

omnigentx (commented):
GitNexus PR review — **Risk: LOW**. Verified against the source, not the diff alone; 44 gateway tests pass locally.

**Verified OK:**
- `run()` registers the menu **once** — `_register_command_menu()` sits before the `while` poll loop, so it does not re-fire per poll/reconnect (matches the comment).
- `_COMMANDS` keys (`new/agent/whoami/help`) are valid Telegram command names (lowercase, ≤32 chars, 4 commands ≤ the 100 limit) — no `BOT_COMMANDS_TOO_MUCH` handling needed.
- `_call(method, body=None)` matches the call sites (`deleteMyCommands` with no body) and the test's `fake_call`.
- Frontend revert-on-failure re-renders: `rows = reactive({})`, and `v-model` flips `r.enabled` before `@change` fires, so `r.enabled = !r.enabled` in the catch reverts the right way.
- Zalo stays `registers_commands=False` (+ regression test), grounded in the Zalo Bot API docs — no 400 per startup.

**Recommendation: APPROVE.** One optional nit inline; nothing blocking.

omnigentx:
Addressed in the latest commit:

**[NIT] In-flight guard on the toggle** — `toggleEnabled` now early-returns if a persist is already running, sets `r.saving` for the duration, and the checkbox is `:disabled="r.saving"`, so rapid flips can't fire overlapping `/settings/bulk` writes or double `pollStatusAfterSave`. Frontend build passes.

**Test note** — the revert-on-failure path isn't covered by an automated test: this project has no unit tests for views, and the Playwright e2e serves `dist` (would need a build + a gateways mock for `/settings/bulk` + `/api/gateways`). Verified live instead (persist → gateway stopped). Happy to add a Playwright e2e for the gateways settings if you'd like it as a follow-up.

